PR #32, opened 2019-03-23
dffml: sources: file: Added bz2 support

yashlamba:


yashlamba:
So I added the condition for bz2 to file.py in commit 3e13c99 , I tested the code the same way with iris dataset and it works absolutely fine. But when I wrote tests for the same in test_file.py, it doesn't seem to work:
Tests:
```
    async def test_open_bz2(self):
        source = FakeFileSource('testfile.bz2')
        m_open = mock_open()
        with patch('os.path.exists', return_value=True), \
                patch('builtins.open', m_open):
            await source.open()
            m_open.assert_called_once_with('testfile.bz2', 'rb')
```
And for close:
```
    async def test_close_bz2(self):
        source = FakeFileSource('testfile.bz2')
        m_open = mock_open()
        with patch('builtins.open', m_open):
            await source.close()
            m_open.assert_called_once_with('testfile.bz2', 'wb')
```
When I run the tests:
Screenshot from 2019-03-23 15-06-42

Do you see any possible workarounds? I am confused because it works with an actual test case but not with this.

dffml/source/file.py
```diff
@@ -3,7 +3,7 @@
 import os
 import abc
 import asyncio
-import gzip
+import gzip, bz2
```
pdxjohnny: Please make a import statement for `bz2` on its own line. i.e.:

```python
import gzip
import bz2
```

pdxjohnny:
This is my bad. I did not review your last PR well enough. If you look at 6e3167e you'll see that the test you added was checking for a call to `builtins.open` whereas it should have been checking for a call to `gzip.open` this has been corrected now. For this patch, you'll want to look for `bz2.open`. Thanks!

yashlamba:
Sorry for that but how did that work up until now? I understood the changes needed to make but it seemed to work for gzip.

pdxjohnny:
It worked because `gzip.open` ends up calling `open`, my guess is that `bz2.open` isn't calling `open` until some data get written to it. Therefore the test you tried didn't work.

yashlamba:
I had some git trouble. Sorry for the compiled commits above. Hope it's okay.

> It worked because `gzip.open` ends up calling `open`, my guess is that `bz2.open` isn't calling `open` until some data get written to it. Therefore the test you tried didn't work.

This made so many things clear in my mind. I was perplexed last time myself. I wrote the tests for bz2 now and am really clear about a lot many things now. Thank you.

yashlamba:
Should I modify CHANGELOG.md?

pdxjohnny:
Nice! Thanks!

PR #96, opened 2019-06-11
[WIP] Added Linear Regression Model

yashlamba:
Still needs a lot of work...


yashlamba:
@pdxjohnny, done with accuracy and train, both are working absolutely fine now (discussed in weekly sync)

model/scratch/tests/test_slr.py
```diff
@@ -0,0 +1,88 @@
...
+        async with self.sources as sources, self.features as features, \
+                self.model as model:
+            async with sources() as sctx, model() as mctx:
+                # Test train
+                await mctx.train(sctx, features)
+                # Test accuracy
+                res = await mctx.accuracy(sctx, features)
+                self.assertGreater(res, 0.9)
+                # Test predict
+                res = [repo async for repo in mctx.predict(sctx.repos(),
```
pdxjohnny: Replace everything after line 83 (`# Test predict`) with:


```python
async for repo, prediction, confidence in mctx.predict(sctx.repos(), features):
    correct = FEATURE_DATA[int(Repo.src_url)][1]
    # Comparison of correct to prediction to make sure prediction is within a reasonable range
    self.assertGreater(prediction, correct - (correct * 0.10))
    self.assertLess(prediction, correct + (correct * 0.10))
```

model/scratch/dffml_model_scratch/model/slr.py
```diff
@@ -0,0 +1,120 @@
...
+from dffml.util.entrypoint import entry_point
+from dffml.util.cli.arg import Arg
+
+
+class SLRConfig(ModelConfig, NamedTuple):
+    predict: str
+    directory: str
+
+
+class SLRContext(ModelContext):
```
pdxjohnny: `SLRContext` looks good

model/scratch/dffml_model_scratch/model/slr.py
```diff
@@ -0,0 +1,120 @@
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2019 Intel Corporation
+'''
+Description of what this model does
+'''
+import os
+import abc
+from typing import AsyncIterator, Tuple, Any, List, Optional, NamedTuple, Dict
+from statistics import mean
```
pdxjohnny: Please implement the `mean` function yourself (We want to avoid importing small unmaintained libraries, numpy is fine that's well maintained).

model/scratch/tests/test_slr.py
yashlamba:
```console
dffml/model/scratch/tests/test_slr.py", line 85, in test_context
    correct = FEATURE_DATA[int(Repo.src_url)][1]
TypeError: int() argument must be a string, a bytes-like object or a number, not 'property'
```
I don't understand this, repo.src_url gives a string only...

model/scratch/dffml_model_scratch/model/slr.py
yashlamba: statistics is a part of python standard library, I don't think it will create a problem. I will replace it if you say...

pdxjohnny: Ah yes it is my bad, I checked pypi without checking the standard library :) let's leave it then!

model/scratch/tests/test_slr.py
pdxjohnny: Ah because I accidentally capitalized `Repo`, it should be

```python
correct = FEATURE_DATA[int(repo.src_url)][1]
```

yashlamba: prediction turns to be a coroutine object not the actual value

yashlamba: oh, prediction was coming out to be a coroutine object
I get it. I will change the implementation

model/scratch/dffml_model_scratch/model/slr.py
```diff
@@ -0,0 +1,120 @@
...
+        feature = await self.applicable_features(features)
+        async for repo in repos:
+            feature_data = repo.features(feature)
+            yield repo, self.predict_input(feature_data[feature[0]]), self.regression_line[2]
+
+
+@entry_point('slr')
+class SLR(Model):
+
+    CONTEXT = SLRContext
```
pdxjohnny: For saving an loading from disk, you'll need to implement these methods

```python
class SLR(Model):

    def _filename(self):
        # Create a file path that is specific to the data we are predicting on
        return os.path.join(self.parent.config.directory, hashlib.sha384(self.config.predict.encode()).hexdigest() + '.json')

    async def __aenter__(self) -> "SLRContext":
        filename = self._filename()
        if os.path.isfile(filename):
            # Load from file using open, and json.load. You'll want to make a self.saved property (in __init__ or something)
        return self

        # Save to the file using open, and json.dump. Dump out the self.saved property to the file.
        pass
```

Then in the context, you will want to access `self.parent.saved`, to save and load your regression line. Probably something like this (turn it into a method):

```python
class SLRContext(ModelContext):

    def get_regression_line(self, features: List[str]):
        # Try to get the line, if it doesn't exist in the saved structure return None
        return self.parent.saved.get(hashlib.sha384((''.join(sorted(features))).encode()).hexdigest(), None)

    def set_regression_line(self, features: List[str], slope, constant, accuracy):
        return self.parent.saved[hashlib.sha384((''.join(sorted(features))).encode()).hexdigest()] = (slope, constant, accuracy)
```

> There should probably be a helper method for `hashlib.sha384((''.join(sorted(features))).encode()).hexdigest()` too so if it's changed one place we don't forget to change it the other place

model/scratch/tests/test_slr.py
yashlamba:
```python3
    self.assertGreater(prediction, correct - (correct * 0.10))
    self.assertLess(prediction, correct + (correct * 0.10))
```
switching signs for these two

yashlamba: Everything works fine now.

yashlamba:
@pdxjohnny A funny thing is happening after the last commit, when I check the accuracy, only the first case fails. I guess that is an exception in the data. If you skip 1 repo, all others pass. Should I get rid of it?

model/scratch/dffml_model_scratch/model/slr.py
yashlamba: On It!!!

pdxjohnny:
> @pdxjohnny A funny thing is happening after the last commit, when I check the accuracy, only the first case fails. I guess that is an exception in the data. If you skip 1 repo, all others pass. Should I get rid of it?

Just change the `Salary` data for it so that it works, this is a bogus dataset anyways :)

pdxjohnny:
Or remove it, whatever

model/scratch/setup.py
```diff
@@ -0,0 +1,58 @@
...
+    ]
+
+setup(
+    name='dffml-model-scratch',
+    version=version,
+    description='',
+    long_description=readme,
+    long_description_content_type="text/markdown",
+    author='John Andersen',
+    author_email='john.s.andersen@intel.com',
```
pdxjohnny: You're the author of this one :)

yashlamba: Oops! Forgot to change that...

model/scratch/dffml_model_scratch/model/slr.py
```diff
@@ -0,0 +1,153 @@
...
+    Model wraping scratch API
+    '''
+    def __init__(self, parent):
+        super().__init__(parent)
+        self.xData = np.array([])
+        self.yData = np.array([])
+        self.regression_line = self.get_regression_line()
+
+    def get_regression_line(self):
+        if self.parent.saved.get('regression_line') is None:
```
pdxjohnny: This should just be (`, None` means return None if not set)


```python
def get_regression_line(self):
    return self.parent.saved.get('regression_line', None)
```

model/scratch/dffml_model_scratch/model/slr.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+@entry_point('slr')
+class SLR(Model):
+
+    CONTEXT = SLRContext
+    
+    def __init__(self, config: SLRConfig) -> None:
+        super().__init__(config)
+        self.saved = {'regression_line': None}
```
pdxjohnny: Saved should be an empty `dict`, because the keys within it that we are assigning to will be those hash values of the features (which you havne't added yet, but it will work the same when you do if you make it an empty `dict`)

```python
self.saved = {}
```

yashlamba:
I am able to dump the regression line successfully but it should read it again in the next attempt right? or am I missing something?

model/scratch/dffml_model_scratch/model/slr.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def _filename(self):
+        # Create a file path that is specific to the data we are predicting on
+        return os.path.join(self.config.directory, hashlib.sha384(self.config.predict.encode()).hexdigest() + '.json')
+
+    async def __aenter__(self) -> 'SLRContext':
+        filename = self._filename()
+        if os.path.isfile(filename):
+            with open(filename) as read:
```
pdxjohnny: you should specify the mode with which to `open`

```python
with open(filename, 'r')
```

model/scratch/setup.py
pdxjohnny: Add two more lines here:

```python
    maintainer='John Andersen',
    maintainer_email='john.s.andersen@intel.com',
```

Since you authored the package, but I'll be the one managing update to the package repository

model/scratch/dffml_model_scratch/model/slr.py
yashlamba: but features won't include regression_line? How will I save that? (I haven't used hashlib before so a bit confused)

yashlamba:
> > > I am able to dump the regression line successfully but it should read it again in the next attempt right? or am I missing something?
> 
> > It should, my terminal is broken right now... I'll re-boot and try to see if I can figure out what's up, try making these changes in the mean-time
> 
> Looks like this works for me, seems to be saving and loading

Weird... I'll check

model/scratch/dffml_model_scratch/model/slr.py
```diff
@@ -0,0 +1,153 @@
...
+        regression_line = [m*x + b for x in xs]
+        accuracy = await self.coeff_of_deter(ys, regression_line)
+        await self.set_regression_line(m, b, accuracy)
+        return (m, b, accuracy)
+
+    async def predict_input(self, x):
+        prediction = self.regression_line[0]*x+self.regression_line[1]
+        return prediction
+
+    async def train(self, sources: Sources, features: Features):
```
pdxjohnny: Since I haven't had time to take out `classifications` yet from the base class, `Model` you'll need to add

```python
, _classifications):
```

to the end of each method: `train`, `accuracy`, `predict`

pdxjohnny:
Here's how to test via CLI

```console
cat > test.csv <<EOF
YearsExperience,Salary
1.1,42393.0
1.3,49255.0
1.5,40781.0
2.0,46575.0
2.2,42941.0
2.9,59692.0
3.0,63200.0
3.2,57495.0
3.2,67495.0
3.7,60239.0
3.9,66268.0
4.0,58844.0
4.0,60007.0
4.1,60131.0
4.5,64161.0
4.9,70988.0
5.1,69079.0
5.3,86138.0
5.9,84413.0
6.0,96990.0
6.8,94788.0
7.1,101323.0
7.9,104352.0
8.2,116862.0
8.7,112481.0
9.0,108632.0
9.5,120019.0
9.6,115685.0
10.3,125441.0
10.5,124922.0
EOF

dffml train \
  -model slr \
  -classifications nil \
  -sources input=csv \
  -log debug \
  -source-filename test.csv \
  -features def:YearsExperience:float:1 \
  -model-predict Salary

cat ~/.cache/dffml/scratch/c4e0c0bb47e180f1118c13da8f01caf8c0408b8aeaf851997ab8dbf03b1440a5f3f1cc6bf52a60c3adb3e22c85f73dc6.json && echo

dffml accuracy \
  -model slr \
  -classifications nil \
  -sources input=csv \
  -log debug \
  -source-filename test.csv \
  -features def:YearsExperience:float:1 \
  -model-predict Salary
```

model/scratch/dffml_model_scratch/model/slr.py
pdxjohnny: So we load / save from a filename based on the feature name we are predicting on. Within SLRContext the `train`, `accuracy`, and `predict` methods create and user regression lines which are based on the feature(s) they were told to use. Therefore, within `self.saved` we want to create a mapping of `features -> regression_line_applicable_to_these_features`.

By using hashlib we know we're always going to get a nice string of hex which we'll have no trouble saving and loading with regards to character encodings.

So we want to take the features that we are training, predicting, accuracy accessing on and save and load from `self.saved` using the hash of those feature names (the one feature in this case). The hash of a value is always going to be the same everytime you hash the same input value, so this is how we'll store the regression lines in saved, by the unique hash associated with the feature the line was trained to predict.

Does that makes sense? Also, check out the documentation for hashlib, play around with it and then you'll understand better how we're using it. https://docs.python.org/3/library/hashlib.html

Essentially, when you call `update` on the hash object you are having it hash more input data. you then call `digest/hexdigest` to get the resulting hash. Imagine you had a large file, and you wanted to calculate a unique value which maps only to those specific contents of that file, change the contents, new hash. You could do this by reading it line by line and passing each line to digest.

model/scratch/dffml_model_scratch/model/slr.py
```diff
@@ -74,11 +75,12 @@ def get_regression_line(self):
         prediction = self.regression_line[0]*x+self.regression_line[1]
         return prediction
 
-    async def train(self, sources: Sources, features: Features):
+    async def train(self, sources: Sources, features: Features, _classifications):
         '''
         Train using repos as the data to learn from.
         '''
         feature = await self.applicable_features(features)
+        self.key = hashlib.sha384((''.join(sorted(feature))).encode()).hexdigest()
```
pdxjohnny: Try following the pattern that I set out with `get_regression_line` and `set_regression_line` please, multiple of these methods could be called at the same time and that would create problems if you use `self.key` like this

yashlamba: should I use a helper method then? or the way you implemented

yashlamba: Any suggestions on how would I pass features on function calls? __init__ and best_fit_line method. Latter is understandable but I think there might be a better...

pdxjohnny: Try this:

https://github.com/intel/dffml/pull/96#discussion_r293017252

yashlamba: I did, but function calls are in __init__ where I don't have access to features

model/scratch/dffml_model_scratch/model/slr.py
```diff
@@ -0,0 +1,153 @@
...
+class SLRContext(ModelContext):
+    '''
+    Model wraping scratch API
+    '''
+    def __init__(self, parent):
+        super().__init__(parent)
+        self.xData = np.array([])
+        self.yData = np.array([])
+        self.key = None
+        self.regression_line = self.get_regression_line()
```
pdxjohnny: You should remove `self.key` and `self.regression_line`, the methods `get_regression_line` and `set_regression_line` are [setters and getters](https://guide.freecodecamp.org/java/getters-and-setters/) which is a concept allowing us to remove usages of variables and replace them with methods which give us greater control.

model/scratch/dffml_model_scratch/model/slr.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def get_regression_line(self):
+        return self.parent.saved.get(self.key, None)
+
+    async def set_regression_line(self, slope, constant, accuracy):
```
pdxjohnny: This doesn't have to be `async`

model/scratch/dffml_model_scratch/model/slr.py
rahulunair: can't we use [property](https://docs.python.org/2/library/functions.html#property) like:

```python
@property
def regression_line(self):
    pass

# if there is a need for a setter 
@regression_line.setter
def regression_line(self, val)
    pass
instead of java like setter and getter to be more pythonic?

File: model/scratch/dffml_model_scratch/model/slr.py [hunk 13]
```diff
@@ -0,0 +1,153 @@
...
+# Copyright (c) 2019 Intel Corporation
+'''
+Description of what this model does
+'''
+import os
+import abc
+import hashlib
+import json
+from typing import AsyncIterator, Tuple, Any, List, Optional, NamedTuple, Dict
+from statistics import mean
```
rahulunair: Recommend using numpy.average/np.mean for performance vs statistics.mean

model/scratch/dffml_model_scratch/model/slr.py
```diff
@@ -0,0 +1,153 @@
...
+        for feature in features:
+            if feature.dtype() != int and feature.dtype() != float:
+                raise ValueError("Simple Linear Regression only supports int or float feature")
+            if feature.length() != 1:
+                raise ValueError("Simple LR only supports single values (non-matrix / array)")
+            usable.append(feature.NAME)
+        return usable
+
+    async def best_fit_line(self, xs, ys):
+        m = ((mean(xs)*mean(ys)) - mean(xs*ys))/((mean(xs)*mean(xs)) - (mean(xs*xs)))
```
rahulunair: perf nit: assign temp vars to function calls and limit function call , matters when you scale and improves readability
```python
mean_xs = np.mean(xs)
mean_ys = np.mean(ys)
m = (mean_xs * mean_ys - np.mean(xs * ys))/(mean_xs ** 2 - np.mean(xs * xs))
b = mean_ys  - m * mean_xs
```

rahulunair: you could even use [np.polyfit](https://docs.scipy.org/doc/numpy/reference/generated/numpy.polyfit.html) to find the intercept and bias, but if @pdxjohnny  wants the implementation to be vanilla without using things like `polyfit`, let's make sure everything is vectorized

model/scratch/dffml_model_scratch/model/slr.py
pdxjohnny: I was hoping we could, but since this is dependent on `features` we can't with the current implementation. Which is now leading me to think. If the `Model` context is always dependent on the `features` maybe we should remove `features` from the arguments to `train`, `accuracy`, and `predict` and we'd end up with a cleaner API.

I think that this is the direction we will go: #101

model/scratch/dffml_model_scratch/model/slr.py
```diff
@@ -0,0 +1,153 @@
...
+        b = mean(ys) - (m*mean(xs))
+        regression_line = [m*x + b for x in xs]
```
rahulunair: Can't we vectorize this, using something similar to?
```python
def linear_eq(x):
    return m * x + b
vectorized_linear_eq = np.vectorize(linear_eq)
regression_line = vectorized_linear_eq(xs)
```

yashlamba: I passed this because it was just simple linear regression (only 2 variables), will it make any difference if I use this?

yashlamba:
I have made most of the requested changes, should I wait for https://github.com/intel/dffml/pull/96#discussion_r295557189 or implement it 
https://github.com/intel/dffml/pull/96#discussion_r295545826 this way?

Issue #48, opened 2019-04-11
source: csv: Enable setting repo src_url


pdxjohnny:
`CSVSource` needs to be able to set a `Repo`s `src_url` to a specified column name.

pdxjohnny:
Probably need a new `Arg` in `args()` and use it in `config()`

PR #102, opened 2019-06-20
source: csv: Enable setting repo src_url using CSVSourceConfig

sudharsana-kjl:
This is a fix for issue #48. Here I've included CSVSourceConfig separately so that FileSourceConfig remains undisturbed with respect to addition of ```key``` parameter.

### DIFF
--- a/dffml/source/csv.py
+++ b/dffml/source/csv.py
@@ -32,6 +32,16 @@ class CSVSource(FileSource, MemorySource):
 
     @classmethod
     def args(cls, args, *above) -> Dict[str, Arg]:
+        cls.config_set(args, above, "filename", Arg())
+        cls.config_set(
+            args,
+            above,
+            "readonly",
+            Arg(type=bool, action="store_true", default=False),
+        )
+        cls.config_set(
+            args, above, "label", Arg(type=str, default="unlabeled")
+        )
         cls.config_set(args, above, "key", Arg(type=str, default=None))
         return args
 
--- a/dffml/source/csv.py
+++ b/dffml/source/csv.py
@@ -8,9 +8,18 @@ import ast
 
 from ..repo import Repo
 from .memory import MemorySource
-from .file import FileSource
+from .file import FileSource, FileSourceConfig
+from typing import NamedTuple, Dict
+from ..util.cli.arg import Arg
 
 csv.register_dialect("strip", skipinitialspace=True)
+
+
+class CSVSourceConfig(FileSourceConfig, NamedTuple):
+    filename: str
+    label: str = "unlabeled"
+    readonly: bool = False
+    key: str = None
 
 
 class CSVSource(FileSource, MemorySource):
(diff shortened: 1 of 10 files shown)


dffml/source/csv.py
```diff
@@ -21,6 +30,20 @@ class CSVSource(FileSource, MemorySource):
     # Headers we've added to track data other than feature data for a repo
     CSV_HEADERS = ["prediction", "confidence", "classification"]
 
+    @classmethod
+    def args(cls, args, *above) -> Dict[str, Arg]:
+        cls.config_set(args, above, "key", Arg(type=str, default=None))
```
pdxjohnny: Don't forget to `config_set` all the variables

tests/source/test_csv.py
```diff
@@ -3,17 +3,66 @@
 import unittest
 
 from dffml.source.file import FileSourceConfig
-from dffml.source.csv import CSVSource
+from dffml.source.csv import CSVSource, CSVSourceConfig
 from dffml.util.testing.source import FileSourceTest
 from dffml.util.asynctestcase import AsyncTestCase
+from dffml.repo import Repo
+from dffml.util.cli.arg import parse_unknown
+import tempfile
```
pdxjohnny: Same as the comment on `typing`

tests/source/test_csv.py
```diff
@@ -3,17 +3,66 @@
...
+                "--source-file-readonly",
+            )
+        )
+        self.assertEqual(config.filename, "feedface")
+        self.assertEqual(config.label, "default-label")
+        self.assertEqual(config.key, "SourceURLColumn")
+        self.assertTrue(config.readonly)
+
+    def config(self, filename, label="unlabeled", key=None, readonly=False):
+        return FileSourceConfig(
```
pdxjohnny: CSVSourceConfig?

dffml/source/csv.py
```diff
@@ -8,11 +8,20 @@
 
 from ..repo import Repo
 from .memory import MemorySource
-from .file import FileSource
+from .file import FileSource, FileSourceConfig
+from typing import NamedTuple, Dict
```
pdxjohnny: Let's move this `typing` import out of the `dffml` imports and up to where we import from the rest of the standard library

dffml/source/csv.py
```diff
@@ -45,6 +78,11 @@ class CSVSource(FileSource, MemorySource):
                         repo_data["features"][key] = ast.literal_eval(value)
                     except (SyntaxError, ValueError):
                         repo_data["features"][key] = value
+                if self.config.key is not None and self.config.key == key:
+                    src_url = value
+                if self.config.key is None:
+                    src_url = str(i)
+            i += 1
```
pdxjohnny: i += 1 needs to be in the loop

tests/source/test_csv.py
sudharsana-kjl: Sorry :sweat_smile:

dffml/source/csv.py
sudharsana-kjl: Oh, i think it has to outside the loop. Because inside the loop we are looking at only one Repo's key, value pairs.

sudharsana-kjl: i+=1 still happens inside the main loop which runs for all the repos. It gets incremented only after we assign it to a particular repo. If i move it inside the inner loop as well, then it has to be incremented for each key, value pair from the data in a row. So this should be added outside the inner loop but within the main loop.

pdxjohnny: Ah yes my bad!

tests/source/test_csv.py
```diff
@@ -1,19 +1,68 @@
...
+                "SourceURLColumn",
+                "--source-file-readonly",
+            )
+        )
+        self.assertEqual(config.filename, "feedface")
+        self.assertEqual(config.label, "default-label")
+        self.assertEqual(config.key, "SourceURLColumn")
+        self.assertTrue(config.readonly)
+
+    def config(self, filename, label="unlabeled", key=None, readonly=False):
```
pdxjohnny: Looks like this `config()`  method is never called, can you please remove it? https://codecov.io/gh/intel/dffml/pull/102/src/tests/source/test_csv.py#L51

sudharsana-kjl: Done :+1:

PR #144, opened 2019-07-26
model: Added scikit linear regression model

yashlamba:


model/scikit/.coveragerc
```diff
@@ -0,0 +1,13 @@
+[run]
+source =
+    REPLACE_IMPORT_PACKAGE_NAME
```
pdxjohnny: This needs to be changed, (due to #146)

model/scikit/LICENSE
```diff
@@ -0,0 +1,21 @@
+Copyright (c) 2017-2019 nucleon
```
pdxjohnny: Who is `nucleon`?

pdxjohnny: Since this is going into the main repo, it's all under Intel, similarly to

https://github.com/google/oss-fuzz/blob/master/projects/boost/build.sh#L2

So this should be

```
Copyright (c) 2019 Intel
```

PR #145, opened 2019-07-29
WIP: source: HDFS 

sudharsana-kjl:
This PR hopes to add HDFS source.


tests/source/test_hdfs.py
```diff
@@ -0,0 +1,22 @@
+from dffml.source.hdfs_source import HDFSSourceConfig, HDFSSource
+
+from dffml.util.testing.source import FileSourceTest
+from dffml.util.asynctestcase import AsyncTestCase
+from dffml.source.csv import CSVSource, CSVSourceConfig
+
+
+class TestHDFSSource(FileSourceTest, AsyncTestCase):
```
pdxjohnny: `FileSourceTest` should be `SourceTest`

dffml/source/hdfs_source.py
```diff
@@ -0,0 +1,58 @@
...
+from dffml.util.cli.arg import Arg
+from dffml.util.entrypoint import entry_point
+
+
+class HDFSSourceConfig(BaseConfig, NamedTuple):
+    host: str
+    port: str
+    user: str
+    filepath: str
+    source: BaseSource.load
```
pdxjohnny: `source: FileSource`

dffml/source/hdfs_source.py
```diff
@@ -0,0 +1,58 @@
...
+@entry_point("dffml.source")
+class HDFSSource(BaseSource):
+
+    CONTEXT = BaseSourceContext
+
+    async def __aenter__(self) -> "BaseSource":
+        self.client = InsecureClient(
+            "http://" + self.config.host + ":" + self.config.port,
+            user="hadoopuser",
+        )
```
pdxjohnny:
```python


async def new_open(source):
    with self.client.read(self.config.filepath, encoding="utf-8") as fd:
        await source.load_fd(fd)
self.config.source._open = new_open
```

Also, give it a new close method, which will be called when we exit from the filesource's context.

To enter the context, we usually do the `async with` statement:

```python
async with self.config.source as source: # this is where new_open get called
    # At the end of this with block new_close would get called, but we don't want to call it until the __aexit__ method of HDFSSource
```

You'll need to do something like this, also make sure to modify `__aexit__` similarly.

https://github.com/intel/dffml/blob/927fb736dd886b4976616f917633696dd3799e49/dffml/df/memory.py#L780-L792

You'll use

`self.source = await self.__stack.enter_async_context(self.config.source)`

dffml/source/hdfs_source.py
```diff
@@ -0,0 +1,58 @@
...
+        with self.client.read(self.config.filepath, encoding="utf-8") as fd:
+            await self.config.source.load_fd(fd)
+        return self
+
+    @classmethod
+    def args(cls, args, *above) -> Dict[str, Arg]:
+        cls.config_set(args, above, "host", Arg(type=str))
+        cls.config_set(args, above, "port", Arg(type=str))
```
pdxjohnny: `type=int`

.ci/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+
+# download and extract hadoop, set JAVA_HOME in hadoop-env.sh, update path
+RUN \
+  wget http://apache.mirrors.tds.net/hadoop/common/hadoop-3.1.2/hadoop-3.1.2.tar.gz && \
+  tar -xzf hadoop-3.1.2.tar.gz && \
+  rm -rf hadoop-3.1.2.tar.gz && \
+  mv hadoop-3.1.2 /usr/local && \
+  ln -sf /usr/local/hadoop-3.1.2/ /usr/local/hadoop
+
+CMD cat hadoop_config >> ~/.bashrc
```
pdxjohnny: You need to `COPY` hadoop_config into the container first

.ci/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+CMD bash cat "" >> ~/.ssh/authorized_keys \
+    chmod 600 ~/.ssh/authorized_keys
+
+## ssh key gen
+RUN \
+    ssh-keygen -t rsa -P "" -f ~/.ssh/id_rsa
+
+CMD bash cat ~/.ssh/id_rsa.pub >> ~/.ssh/authorized_keys \
+    ssh-copy-id -i ~/.ssh/id_rsa.pub localhost \
+    ssh localhost
```
pdxjohnny: This won't work because the ssh server isn't running. You can safely delete these steps

.ci/Dockerfile
sudharsana-kjl: I've included the following line:
```
ADD configs/hadoop_config /usr/local/hadoop
CMD cat /usr/local/hadoop/hadoop_config >> ~/.bashrc
```
Is this fine?

.ci/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+##RUN \
+##    mkdir ~/.ssh \
+##    chmod 700 ~/.ssh
+
+CMD bash cat "" >> ~/.ssh/authorized_keys \
+    chmod 600 ~/.ssh/authorized_keys
+
+## ssh key gen
+RUN \
+    ssh-keygen -t rsa -P "" -f ~/.ssh/id_rsa
```
pdxjohnny: You don't need this, since you don't need to create an ssh key for the root account within the container

.ci/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+CMD bash ~/.bashrc
+  ##echo "export JAVA_HOME=$JAVA_HOME" >> $HADOOP_HOME/etc/hadoop/hadoop-env.sh && \
+  ##echo "PATH=$PATH:$HADOOP_HOME/bin" >> ~/.bashrc
+
+##ssh setup
+##RUN \
+##    mkdir ~/.ssh \
+##    chmod 700 ~/.ssh
+
+CMD bash cat "" >> ~/.ssh/authorized_keys \
```
pdxjohnny: You can remove this

.ci/Dockerfile
sudharsana-kjl: Oh, so there's no need to fetch the authorized keys?

sudharsana-kjl: Or I should just remove ```ssh localhost``` ?

.ci/Dockerfile
sudharsana-kjl: okay :+1:

.ci/Dockerfile
sudharsana-kjl: okay :+1:

.ci/Dockerfile
sudharsana-kjl: Got it :+1:

.ci/configs/ssh_config
```diff
@@ -0,0 +1,3 @@
+Host *
+  UserKnownHostsFile /dev/null 
+  StrictHostKeyChecking no
```
pdxjohnny: You should remove this file

.ci/start-hadoop.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
```
pdxjohnny: You probably want to `source /usr/local/hadoop/hadoop_config`

dffml/source/hdfs_source.py
pdxjohnny: Implement __call__ method for HDFS source which does: `return self.source()`

pdxjohnny: https://github.com/intel/dffml/blob/eb9b6002d6fb4df83f79dd57ae8992fb372413ad/dffml/source/source.py#L60-L61

sudharsana-kjl: Done :+1:

pdxjohnny:
```diff
diff --git a/dffml/source/hdfs_source.py b/dffml/source/hdfs_source.py
index 9f90f62..2040b5f 100644
--- a/dffml/source/hdfs_source.py
+++ b/dffml/source/hdfs_source.py
@@ -37,12 +37,23 @@ class HDFSSource(BaseSource):
             "http://" + self.config.host + ":" + self.config.port,
             user="hadoopuser",
         )
-        self.config.source._open = await self.new_open()
-        self.config.source._close = await self.new_close()
+        self.config.source._open = self.new_open.__get__(self.config.source, self.config.source.__class__)
+        self.config.source._close = self.new_close.__get__(self.config.source, self.config.source.__class__)
+        # TODO Create AsyncExitStack and call it's __aenter__ method
+        self.__stack = ...
+        # Enter the context of the source
+        # Same as:
+        # async with self.config.source as self.source:
+        self.source = self.__stack.async_enter_context(self.config.source)
         return self
 
+    # TODO add __aexit__ method
+
     def __call__(self) -> BaseSourceContext:
-        return self.config.source()
+        # Call the source
+        # Same as:
+        # async with self.source() as source_context:
+        return self.source()
 
     @classmethod
     def args(cls, args, *above) -> Dict[str, Arg]:
```

PR #190, opened 2019-09-06
Added docs for scikit models

yashlamba:
@pdxjohnny I'm not sure if I did the docs.py part right, I can make changes as required.


scripts/docs.py
```diff
@@ -135,11 +136,13 @@ def format_op(op):
 
 
 def gen_docs(entrypoint: str, modules: List[str], maintenance: str = "Core"):
-    per_module = {name: [] for name in modules}
+    per_module = {name: [None, []] for name in modules}
```
pdxjohnny: I was going to tell you to try to avoid importing the package directly, but I didn't know how to explain so I just did it to make sure it worked and then it was done so I pushed.

pdxjohnny:
Nice! This looks great I built the site and this makes usage very clear. I think we need to add an example usage of the classification as well, would that make sense or what do you think?

yashlamba:
I'll add those examples + I'll replace the scikitlr example with the one we have in README  of scikit. I was going through changed files, why are individual model args still defined in the docs? Do we need that or something?

docs/plugins/dffml_model.rst
```diff
@@ -287,26 +352,590 @@ Linear Regression Model implemented using scikit. Models are saved under the
...
+
+  - default: -1
+
+- decision_function_shape: String
+
+  - default: ovr
+
+- random_state: BaseConfigurable.parser_helper
+
+  - default: None
```
yashlamba: Why does this exist?

pdxjohnny: Well, docs.py needs to generate the args for everything. Those definitions are weird https://github.com/intel/dffml/issues/194. We could make it so that the docs script has a list of things it shouldn't print out args() for. That would be good.

yashlamba: It was working in 3ed5298, we can change the condition to a list maybe here https://github.com/intel/dffml/pull/190/commits/3ed529811d2a3a1558365b2dd52e9a00c195b853#diff-69087563aad9b14661594d9bd805fcd2R173-R177

pdxjohnny: Ah, yes, I didn't remember to change the if statement as well. Maybe we could add

```python
"\n\n".join(docs) if module_name not in list_of_module_names_not_to_include
```

and load `list_of_module_names_not_to_include` from a file, similarly to how we load from `scripts/care`. Or you could make it so we use [configparser](https://docs.python.org/3/library/configparser.html) to read the docs.py config from a file and put the list in there.

pdxjohnny: so, consolidate the scripts/care and the list of modules not to write args for into one config file that gets read with configparser.

yashlamba:
@pdxjohnny I have updated the script but haven't pushed the doc files yet as I was having a lot of warnings regarding other packages. Can you please try the script and push the updated docs?

PR #219, opened 2019-10-09
Bug rename init arg in DNNClassifierModelContext

DmitryPoliuha:
Was renamed DNNClassifierModelContext first init arg from config to features


pdxjohnny:
Thanks!

PR #225, opened 2019-10-17
WIP: Added Model not trained error

yashlamba:


model/scikit/dffml_model_scikit/scikit_base.py
```diff
@@ -60,7 +61,7 @@ def _filename(self):
         return self
 
     async def __aexit__(self, exc_type, exc_value, traceback):
-        joblib.dump(self.clf, self._filename())
+        pass
```
pdxjohnny: Was this removed to test? I assume it'll be back when it's not WIP right? I may be missing something

yashlamba: Nope, it was actually a bug. We don't want to save an untrained model but it does the same. It is fulfilled by https://github.com/intel/dffml/blob/1d4c04b2e4790a3533eb92469cb904c80782d27b/model/scikit/dffml_model_scikit/scikit_base.py#L78

Issue #232, opened 2019-10-24
ci: Whitespace checker broken


pdxjohnny:
`.ci/run.sh` whitespace checking function is broken. It should exit non-zero if there is any trailing whitespace in the repo

beinganonymous:
@pdxjohnny i would like to work on this.

pdxjohnny:
@beinganonymous Awesome! Thanks!

drigio:
Hey @pdxjohnny initial looks at the code in ``.ci/run.sh`` looks good. Also I took a look at previous builds on travis and it has a good history. Running the code locally on pc (I ran only the ``run_whitespace()`` function) works properly. When I add an extra whitespace line in any file it duly shows that there are trailing whitespaces.

Checking the exit code by using ``echo $?`` returns 1 i.e. non-zero as well. 
What you think @beinganonymous ? 
Are you getting some different output @pdxjohnny ? Because I guess Linux distribution and its accompanying toolset (find, grep) etc may matter too.

I am running a distribution that inherits Ubuntu 18.04 LTS.

pdxjohnny:
Huh, maybe its not broken then. I'll have to check last time I tried it didn't exit 1, which is why I made the issue. Thanks for validating @drigio! I appreciate it, probably something wrong with my machine then. Let me see if I can reproduce by own bug lol

beinganonymous:
@drigio when i am checking the exit code using ``echo $?`` it returns 0.
I am using Ubuntu 16.04 LTS.

beinganonymous:
@pdxjohnny are you expecting trailing whitespace as a space bar or tab at end or as a new line at the end ?

beinganonymous:
Hey @pdxjohnny  your ``run_whitespace()`` function work properly with ``Ubuntu 18.04``, but i guess it has issues with ``Ubuntu 16.04``.
I have created this script which work fine with Ubuntu 18.04 and 16.04.
```
    for file in `find $1 -type f -name "*.py"`;
    do
        nlines=`tail -n 1 $file | grep '^$' | wc -l`
        if [ $nlines -eq 1 ]
                then echo $file
        fi
     done;
```
Should i send in PR for that?
What you think @pdxjohnny .
What your thoughts about the code @drigio .

beinganonymous:
Hey @pdxjohnny  your ``run_whitespace()`` function work properly with ``Ubuntu 18.04``, but i guess it has issues with ``Ubuntu 16.04``.
[...]

PR #237, opened 2019-10-31
bug : run.sh : WhiteSpace Checker Solved.

beinganonymous:
This PR closes issue #232 

``ci/run.sh`` whitespace checking function was not working fine, so added some changes to it which now work fine, it exits non-zero value if trailing whitespace is found.

I have added whitespaces to two file ``base.py`` and ``cli.py`` .
After running script it shows filename with trailing whitespaces and exit with non-zero value.

image

I am using Ubuntu 16.04 LTS.

### DIFF
--- a/.ci/run.sh
+++ b/.ci/run.sh
@@ -74,8 +74,13 @@ function run_whitespace() {
     rm -f "$whitespace"
   }
   trap rmtempfile EXIT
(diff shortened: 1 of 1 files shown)


pdxjohnny:
I'm sorry but this failed the CI. We've since moved to GitHub Actions where the whitespace checker seems to be working (70ce150). Thank you for giving this a shot though.

Issue #254, opened 2019-12-05
docs: usage: integration: Make prediction using operation


pdxjohnny:
In the Integrating Machine Learning usage example between the [Making a Prediction](https://intel.github.io/dffml/usage/integration.html#making-a-prediction) and [Modifying the Legacy App](https://intel.github.io/dffml/usage/integration.html#modifying-the-legacy-app) headings we need to add a section where we show how to use the Trained Model Operation to make the prediction, rather than the other CLI call.

In the new section we need to go over how we construct the dataflow.

```python
elif action == 'predict':
    today = datetime.now().strftime('%Y-%m-%d %H:%M')
    subprocess.check_call([
        "dffml", "dataflow", "run", "repos", "set",
        "-keys", query['URL'],
        "-repo-def", "URL",
        "-dataflow", os.path.join(os.path.dirname(__file__), "dataflow.yaml"),
        "-sources", "db=demoapp",
        ])
    result = subprocess.check_output([
        'dffml', 'predict', 'repo',
        '-keys', query['URL'],
        '-model', 'tfdnnc',
        '-model-classification', 'maintained',
        '-model-classifications', '0', '1',
        '-sources', 'db=demoapp',
        '-features',
        'def:authors:int:10',
        'def:commits:int:10',
        'def:work:int:10',
        '-log', 'critical',
        '-update'])
    result = json.loads(result)
    cursor.execute("REPLACE INTO status (src_url, maintained) VALUES(%s, %s)",
                   (query['URL'], result[0]['prediction']['value'],))
    cnx.commit()
    print json.dumps(dict(success=True))
```

will turn into:

```python
elif action == 'predict':
    subprocess.check_call([
        "dffml", "dataflow", "run", "repos", "set",
        "-keys", query['URL'],
        "-repo-def", "URL",
        "-dataflow", os.path.join(os.path.dirname(__file__), "prediction_dataflow.yaml"),
        "-sources", "db=demoapp",
        ])
    print json.dumps(dict(success=True))
```

This issue is probably worth spiting into smaller issues. Since off the top of my head I'd say it involves at least these sub-tasks.

- [x] Create a `model_predict` Operation (#215)
  - This operation will do almost the same thing as the `dffml predict repo` CLI command.
  - Done: [model_predict operation](https://github.com/intel/dffml/blob/ba96ad764a1788491307ca93e619f6791bfbf1f0/dffml/operation/model.py#L10-L38)
- [x] Create a `run_dataflow` Operation
[...]

PR #268, opened 2019-12-16
WIP:op:RunDataFlowOnRepo

aghinsa:
operation description given in #254


tests/test_dfrepos.py
```diff
@@ -0,0 +1,123 @@
...
+    async def test_run(self):
+        self.required_plugins("shouldi")
+        stdout = io.StringIO()
+
+        with relative_chdir("..", "..", "examples", "shouldi"):
+            with unittest.mock.patch("sys.stdout.buffer.write") as write:
+                await Develop.cli("export", "shouldi.cli:DATAFLOW")
+            shouldi_dataflow = DataFlow._fromdict(
+                **json.loads(write.call_args[0][0])
+            )
```
pdxjohnny: Using the `shouldi` dataflow is a great test. However I think we probably want to avoid it here since it does network requests and slows down our already long tests for the main repo. This will also avoid the event that the number of issues might change.

It would be good to take the test dataflow in `tests/test_df.py` and make it a global variable. Then you can import it easily from this file and use that as the test dataflow.

https://github.com/intel/dffml/blob/85cd82379042d543fb280163fee68b0043cc94da/tests/test_df.py#L181-L184

aghinsa:
@pdxjohnny made the changes :+1:

PR #274, opened 2019-12-25
WIP: usage: integration: Simplify example

aghinsa:
operation desciption in #254 .

### DIFF
--- a/dffml/df/types.py
+++ b/dffml/df/types.py
@@ -381,7 +381,11 @@ class Input(object):
         return repr(self)
 
     def export(self):
-        return dict(value=self.value, definition=self.definition.export())
+        return dict(
+            value=self.value,
+            definition=self.definition.export(),
+            origin=self.origin,
+        )
 
     @classmethod
     def _fromdict(cls, **kwargs):
--- a/dffml/feature/feature.py
+++ b/dffml/feature/feature.py
@@ -58,7 +58,7 @@ class Feature(abc.ABC):
             dtype = tempvar[1]
             length = parser_helper(tempvar[2])
         if isinstance(dtype, str):
-            dtype = self.convert_dtype(tempvar[1])
+            dtype = self.convert_dtype(dtype)
         self.dtype = dtype
         self.length = length
         self.name = name
--- a/examples/maintained/make_prediction_dataflow.py
+++ b/examples/maintained/make_prediction_dataflow.py
@@ -1,0 +1,235 @@
+import json
+import yaml
+import asyncio
+import copy
+
+
+from dffml import (
+    op,
+    DataFlow,
+    Input,
+    InputFlow,
+    MemoryOrchestrator,
+    Feature,
+    Definition,
+    Features,
+    GetSingle,
+    run_dataflow,
+    RunDataFlowConfig,
+)
+
+from dffml.configloader.configloader import ConfigLoaders
+from dffml.db.sqlite import SqliteDatabase, SqliteDatabaseConfig
+
+from dffml.df.base import op
+from dffml.operation.mapping import (
+    mapping_extract_value,
+    create_mapping,
+    mapping_merge,
+    group_mappings,
+)
+from dffml.operation.model import model_predict, ModelPredictConfig
+from dffml.operation.db import (
+    DatabaseQueryConfig,
+    db_query_insert,
+)
+
+from dffml_source_mysql.db import (
+    MySQLDatabase,
+    MySQLDatabaseConfig,
+)
+
+from dffml_model_tensorflow.dnnc import (
+    DNNClassifierModel,
+    DNNClassifierModelConfig,
+)
+
+from feature.git.dffml_feature_git.feature.definitions import *
+from dffml.operation.output import GroupBy
+
+
+async def main():
+    subflow_yaml = "./cgi-bin/dataflow.yaml"
+    async with ConfigLoaders() as cfgl:
+        _, subflow = await cfgl.load_file(filepath=subflow_yaml)
+    subflow = DataFlow._fromdict(**subflow)
+
+    # We want to create a dataflow which runs the
+    # subflow and use the outputs of the subflow as
+    # features to run inference on a trained model,
+    # and add the prediction of the model to a database.
+
+    prediction_df = DataFlow(
+        operations={
+            run_dataflow.op.name: run_dataflow,
+            GetSingle.op.name: GetSingle,
+            model_predict.op.name: model_predict,
+            group_mappings.op.name: group_mappings,
+            db_query_insert.op.name: db_query_insert,
+            "mapping_extract_value": mapping_extract_value.op,
+            "create_key_mapping": create_mapping.op,
+            "create_maintained_mapping": create_mapping.op,
+            "create_insert_data": mapping_merge,
+        }
+    )
+
+    prediction_df.configs[db_query_insert.op.name] = DatabaseQueryConfig(
+        database=MySQLDatabase(
+            MySQLDatabaseConfig(
+                host="127.0.0.1",
+                port=3306,
+                user="user",
+                password="pass",
+                db="db",
+                ca=None,
+            )
+        )
+    )
+    prediction_df.configs[run_dataflow.op.name] = RunDataFlowConfig(
+        dataflow=subflow
+    )
+    prediction_df.configs[model_predict.op.name] = ModelPredictConfig(
+        model=DNNClassifierModel(
+            DNNClassifierModelConfig(
+                predict=Feature("maintained", int, 1),
+                classifications=[0, 1],
+                features=Features(
+                    Feature("authors", int, 10),
+                    Feature("commits", int, 10),
+                    Feature("work", int, 10),
+                ),
+            )
+        )
+    )
+    # Modifying run_dataflow to take URL as input and output the features
+    prediction_df.operations[run_dataflow.op.name] = run_dataflow.op._replace(
+        inputs={"URL": URL},
+        outputs={
+            "authors": author_count,
+            "commits": commit_count,
+            "work": work_spread,
+        },
+        expand=[],
+    )
+    # Modyfing group_mapping to gather the features (results of the subflow)
+    prediction_df.operations[
+        group_mappings.op.name
+    ] = group_mappings.op._replace(
+        inputs={
+            "authors": author_count,
+            "commits": commit_count,
+            "work": work_spread,
+        },
+        outputs={"results": model_predict.op.inputs["features"]},
+    )
+
+    # For inserting the prediction to the database,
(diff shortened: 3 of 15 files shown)


pdxjohnny:
Also, see the TODO I added, please address that

aghinsa:
> This is looking great. Nice work.
> 
> I made a couple tweaks.
> 
>     * Reformatted with black
> 
>     * Removed the `db` from the sqlite filenames (since they're already in the `db` directories)
> 
>     * Moved the test in with the rest of the tests
> 
>     * Moved `make_condition_expression` into `sqlite.py` since it's specific to sql expressions.
> 
>     * Changed it so that lookup is an iterator so other databases can asynchronously yield results as they come in.
> 
>     * Made it so that lookup returns rows which are `dict`s so that they contain the column names in them. Hopefully this will make easy to pass the rows around to other operations which require `dict` values
> 
> 
> A few things that I noticed
> 
>     * We can do `(X OR Y) AND Z` but not `(X AND Y) OR Z`
>       
>       * Correct me if I'm wrong here but it seems just like this is the way `make_condition_expression` is designed to work. Having given it some thought I think if we try to allow for all combinations of things we'll end up with a nightmare. I think if this is the indented functionality then let's document it somewhere. I think it should probably be documented in such a way that it's apparent all other database plugins should follow this (for now, we may change our minds or come up with a better solution later).
> 
>     * If we have two operations which both take the same database we'll need some way of telling them they should use the same instance of `SqliteDatabase`, instead of using too different instances with the same config. This is because if they use different instances then the file will be open twice. This is something we've run into before and solved in sub-par ways (aka json and csv sources).
>       
>       * We need to figure out how we want to define configs which should be used to instantiate a single instance, and then tell other objects to use the single instance. This is something we should discuss next week or the week after. As there will be significant changes required for this. For now though not important, as we'll only need one operation to complete this issue.
> 
> 
> I don't have any wisdom for you one how the operation should work right now unfortunately. I'm swamped at the moment with other things I need to get done. If you want to go ahead and take a stab at it in a similar way as the model_predict operation works. That'd be great. Otherwise I'll have more time next week to advise.

Great!!

* You are right,as of now we can't have expressions of the form `(X AND Y) OR Z`.We could maybe add an option to have a negation on individual expressions (`X`,`Y`,`Z`) but I think it'd be best to leave it as such,atleast for now.
* Sure

aghinsa:
> Also, see the TODO I added, please address that

Done.

dffml/db/sqlite.py
```diff
@@ -0,0 +1,200 @@
...
+                }
+        """
+
+        def _make_condition_expression(conditions):
+            def make_or(lst):
+                val_list = []
+                exp = []
+
+                for cnd in lst:
+                    exp.append(f"({cnd.column} {cnd.operation} ? )")
```
pdxjohnny:
```python
                for cnd in lst:
                    exp.append(f"(`{cnd.column}` {cnd.operation} ? )")
```

dffml/operation/sqlite.py
```diff
@@ -0,0 +1,77 @@
+import inspect
```
pdxjohnny: This file should be named db or something not sqlite specific.

dffml/operation/db.py
```diff
@@ -0,0 +1,189 @@
...
+QUERY_CONDITIONS=Definition(name="query_conditions",primitive="Conditions")
+QUERY_COLS=Definition(name="query_cols",primitive="List[str]")
+QUERY_LOOKUPS=Definition(name="query_lookups", primitive="Dict[str, Any]")
+
+@config
+class SqliteQueryConfig:
+    """
+        query_type : "create","insert","update","remove","lookup"
+    """
+    database : SqliteDatabase
```
pdxjohnny: Since you changed the filename, lets also change all the class, op, etc. names in this to be `db_` rather than `sqlite_` so the config would be `database: BaseDatabase`

dffml/operation/db.py
```diff
@@ -0,0 +1,189 @@
...
+                                        cols=cols,
+                                        conditions=conditions)
+                return {"lookups":[res async for res in result] }
+
+
+
+
+
+
+
```
pdxjohnny: let's remove these blank lines

dffml/operation/db.py
```diff
@@ -0,0 +1,189 @@
...
+    imp_enter={"database": (lambda self: self.config.database)},
+    ctx_enter={"dbctx": (lambda self: self.parent.database())},
+)
+async def sqlite_query_remove(self,*,
+                table_name : str,
+                conditions : Conditions = []
+                ) -> Dict[str, Any]:
+                await self.dbctx.remove(table_name=table_name,
+                                        conditions=conditions)
+                return {"lookups":{}}
```
pdxjohnny: I can't remember, was there a reason we wanted to return lookups from the non-lookup operations still? Or was this just an artifact of the literal translation of spiting the operation into multiple operations?

aghinsa: My bad,copy paste errors.removed it.

examples/maintained/make_prediction_dataflow.py
```diff
@@ -49,30 +47,60 @@
...
+        }
+    ,
+    outputs ={
+        "features" : Definition(
+                name="maintained_features",
+                primitive="Dict[str,Any]"
+                )
+            }
+)
+async def collect_maintained_features(authors,work,commits):
```
pdxjohnny: Great idea with this!

examples/maintained/make_prediction_dataflow.py
```diff
@@ -49,30 +47,60 @@
...
+    return {
+        "features":{
+            "authors":authors,
+            "work":work,
+            "commits":commits
+        }
+    }
+
+#does outputing same defintion cause context to run twice?
```
pdxjohnny: Yes. If you output something that's the same type as your input. Unless you update the `flow` to not have the operation accept inputs from itself.

examples/maintained/make_prediction_dataflow.py
pdxjohnny: This makes me think we might want to add some logic to `DataFlow._fromdict` to use `dffml.util.entrypoint:load` to load operation implementations and add them to `DataFlow.implementations` if the name is in the entrypoint format.

```yaml
  collect_maintained_features:
    inputs:
      authors: author_count
      commits: commit_count
      work: work_spread
    name: make_prediction_dataflow:collect_maintained_features
    outputs:
      features: repo_features
    stage: processing
```

That way users wouldn't have to add to the `setup.py` (or not even create one) for these kinds of one-off operations.

aghinsa: > This makes me think we might want to add some logic to `DataFlow._fromdict` to use `dffml.util.entrypoint:load` to load operation implementations and add them to `DataFlow.implementations` if the name is in the entrypoint format.
> 
> ```yaml
>   collect_maintained_features:
>     inputs:
>       authors: author_count
>       commits: commit_count
>       work: work_spread
>     name: make_prediction_dataflow:collect_maintained_features
>     outputs:
>       features: repo_features
>     stage: processing
> ```
> 
> That way users wouldn't have to add to the `setup.py` (or not even create one) for these kinds of one-off operations.

Yes,i was wondering about this,i'll do that.

pdxjohnny:
- In the tutorial we need to have them install
  - dffml-model-tensorfow
  - dffml-source-mysql
  - dffml-feature-git

- Remove `Pulling from the Database` section

- Change `Making a Prediction` to `Making a single Prediction`

- Add section `Saving Prediction In Database`
  - Export new dataflow to yaml file, include in this section. Include mermaidjs diagram to help with visual explanation.
  - Would be helpful to add example data inputs and outputs from as data moves between operations.

pdxjohnny:
I cleaned up this branch and merged the database abstraction, sqlite and mysql implementations.

The old branch is over here (as a backup, in case I messed anything up): https://github.com/pdxjohnny/dffml/tree/sql_op_backup

dffml/df/types.py
```diff
@@ -629,6 +630,19 @@ def _fromdict(cls, *, linked: bool = False, **kwargs):
...
         }
+
+        # Load operations whose names are in entrypoint format
+        _implementations = {}
+        for _, __op in kwargs["operations"].items():
+            _name = __op.name
+            if _name.count(":") > 0:
+                _implementations[_name] = next(load_entrypoint(__op.name)).imp
+
+        if "implementations" in kwargs:
```
pdxjohnny, line 641: Don't need this anymore

dffml/feature/feature.py
```diff
@@ -52,12 +52,13 @@ class Feature(abc.ABC):
...
         super().__init__()
+        tempvar = None
         if name.count(":") == 2:
             tempvar = name.split(":")
             name = tempvar[0]
             dtype = tempvar[1]
             length = parser_helper(tempvar[2])
-        if isinstance(dtype, str):
+        if isinstance(dtype, str) and tempvar:
             dtype = self.convert_dtype(tempvar[1])
```
pdxjohnny, line 62:
```suggestion
            dtype = self.convert_dtype(dtype)
```

dffml/feature/feature.py
```diff
@@ -87,6 +88,8 @@ def export(self):
 
     @classmethod
     def _fromdict(cls, **kwargs):
+        if isinstance(kwargs["dtype"],str):
```
pdxjohnny, line 91:
```suggestion
```

dffml/feature/feature.py
```diff
@@ -87,6 +88,8 @@ def export(self):
...
+            kwargs["dtype"] = Feature.convert_dtype(kwargs["dtype"])
```
pdxjohnny, line 92:
```suggestion
```

tests/test_prediction_with_query.py
```diff
@@ -0,0 +1,288 @@
...
+                )
+                await ctx.insert(
+                    self.table_name, {"key": "add_op", "value": 0}
+                )
+                await ctx.insert(
+                    self.table_name, {"key": "mult_op", "value": 0}
+                )
+
+    async def test_run(self):
+        test_dataflow = DataFlow(
```
pdxjohnny, line 100: It looks like this dataflow is the same as the one I was suggesting we create using the create command. We need to fully automate the testing of that demo. At this point, it may be best if I take this one from here since you probably need to start focusing on GSoC stuff now. Is that okay with you, or do you think you have time to finish for both? I suspect you do not, but am not going to finish it if you say you want to. You've taken this a long way. At this point all that's left is putting things in `.sh` files and automating it similarly to how the model tutorial was automated.

aghinsa: You can take it up. Thanks!. I have pushed some more changes which I missed earlier, now this should be at a point without errors.

PR #278, opened 2019-12-31
Autoconvert definitions

### LINKED ISSUE #86: df: types: Autoconvert Definitions with spec into their spec
Conversion should happen upon instantiation of an `Input`, if the `value` is a `dict` and the definition has a `spec`, convert the `dict` into and instance of class provided by the `spec` and use that as the `value`.

https://github.com/intel/dffml/blob/01a3deb4bfa01bd1ce8ff690358b46d8b1e037bc/dffml/df/types.py#L253-L278

For testing that this works, modify the git feature operations which use `git_repository` and `git_repository_checked_out`.

https://github.com/intel/dffml/blob/01a3deb4bfa01bd1ce8ff690358b46d8b1e037bc/feature/git/dffml_feature_git/feature/operations.py#L273-L348

`repo["directory"]` will become `repo.directory`, etc. There are more operations towards the top of the file that will need modification. Hold off on changing `clone_git_repo` until we have a working pull request.

Definitions and specs for reference:

https://github.com/intel/dffml/blob/01a3deb4bfa01bd1ce8ff690358b46d8b1e037bc/feature/git/dffml_feature_git/feature/definitions.py#L7-L35

sudharsana-kjl:
Fix #86


dffml/df/types.py
```diff
@@ -269,6 +269,10 @@ def __init__(
         # instance name this Input is intended for.
         if parents is None:
             parents = []
+
+        if isinstance(value, dict) and definition.spec is not None:
+            self.value = definition.spec(**value)
```
pdxjohnny: `value = definition.spec(**value)` :)

pdxjohnny:
How's it going with this? Were you able to get the development environment fixed? [gitpod](https://gitpod.io/#https://github.com/intel/dffml) could be a good option.

sudharsana-kjl:
Sorry John, work has been hectic. I'll give you an update by next call. I'll checkout gitpod +1

pdxjohnny:
No worries! Just wanted to check in. Thanks! :)

dffml/df/types.py
```diff
@@ -269,6 +269,10 @@ def __init__(
         # instance name this Input is intended for.
         if parents is None:
             parents = []
+
+        if isinstance(value, dict) and definition.spec is not None:
+            value = definition.spec(**value)
+        
         self.value = value
```
pdxjohnny: Right before we set `self.value` is where we'll add #349

feature/git/dffml_feature_git/feature/operations.py
```diff
@@ -116,7 +116,7 @@
     outputs={"repo": git_repository_checked_out},
 )
 async def git_repo_checkout(repo: Dict[str, str], commit: str):
-    await check_output("git", "checkout", commit, cwd=repo["directory"])
+    await check_output("git", "checkout", commit, cwd=repo.directory)
     # NOTE Don't modify variables which are mearly references! This will create
     # more permutations than intended.
     checked_out = repo.copy()
```
pdxjohnny:
```
  File "/home/runner/work/dffml/dffml/feature/git/dffml_feature_git/feature/operations.py", line 122, in git_repo_checkout
    checked_out = repo.copy()
AttributeError: 'GitRepoSpec' object has no attribute 'copy'
```

The note here can be removed now that `repo` is an instance of `GitRepoSpec` which is immutable since it's a `NamedTuple`.

We should change this to:

```python
async def git_repo_checkout(repo: Dict[str, str], commit: str):
    await check_output("git", "checkout", commit, cwd=repo.directory)
    return {
        "repo": GitRepoCheckedOutSpec(
            URL=repo.URL,
            directory=repo.directory,
            commit=commit,
        )
    }
```

Issue #272, opened 2019-12-24
model: scikit: Add new scikit based models


pdxjohnny:
@sakshamarora1 is working on adding more models to scikit models

Current issue is that any classes with `__init__` arguments that don’t have default values are not being added as properties to cls.CONFIG.

### Plan
- [x] Focus on figuring out what models have no parameters that have no default values.
- [x] Verify that those models work
- [ ] Identify a few models which are failing
- [x] Get the make_config implemented (#164)
- [x] Change namedtuple(...) calls which create dffml_config into calls to make_config(...)
  - [x] This requires parsing the docstring to identify which types a scikit parameter should be (this can be done at any time)

```python
MAPPING = {
    "string": str,
    "list of dicts": List[dict],
    # ...
}
```

- Make sure we have all the mapping for at least one class so we can start testing this out.

https://github.com/scikit-learn/scikit-learn/blob/bf24c7e3d/sklearn/ensemble/_forest.py#L851-L1013

In the future we will make it so we can dynamically instantiate any scikit model.

This issue is a part of #29

PR #280, opened 2020-01-01
model: scikit: Added New scikit model

0dust:
#272


pdxjohnny:
Very nice! Thank you!

PR #290, opened 2020-01-06
WIP:model :Scikit: Adding Scikit clustering Models

0dust:


tests/integration/test_models.py
```diff
@@ -217,3 +217,107 @@ class TestScikitRegression(IntegrationCLITestCase):
...
+
+class TestScikitClustering(IntegrationCLITestCase):
+    async def test_run(self):
+        self.required_plugins("dffml-model-scikit")
+        # Create the training data
+        train_filename = self.mktempfile() + ".csv"
+        pathlib.Path(train_filename).write_text(
+            inspect.cleandoc(
+                """
+                crim,zn,indus,chas,nox,rm,age,dis,rad,tax,ptratio,b,lstat
```
pdxjohnny: Let's generate a random dataset to use for this. It looks like this might be an existing dataset.

model/scikit/tests/test_scikit.py
```diff
@@ -178,6 +208,58 @@ def tearDownClass(cls):
     [21.20000076, 18.60000038, 8.7, 112481.0],
 ]
 
+FEATURE_DATA_CLUSTERING = [
+    [-9.01904123, 6.44409816, 5.95914173, 6.30718146],
```
pdxjohnny: Just to be sure, this is randomly generated data right? If not let's make sure it is.

model/scikit/dffml_model_scikit/scikit_base.py
```diff
@@ -119,6 +121,94 @@ def _filename(self):
...
+
+        estimator_type = getattr(self.clf, "_estimator_type")
+        if estimator_type is "clusterer":
+            if hasattr(self.clf, "predict"):
+                # inductive clusterer
+                predictor = self.clf.predict
+            else:
+                # transductive clusterer
+                labels = [
+                    (yield label) for label in self.clf.labels_.astype(np.int)
```
unrahul: so, here shouldn't the model be retrained? or is it some form of online learning?

model/scikit/dffml_model_scikit/scikit_base.py
```diff
@@ -119,6 +121,94 @@ def _filename(self):
...
+
+    def _feature_predict_hash(self):
+        params = "".join(
+            [
+                "{}{}".format(k, v)
+                for k, v in self.parent.config._asdict().items()
+                if k not in ["directory", "features"]
+            ]
+        )
+        return hashlib.sha384(
```
unrahul: qq, why is a crypto hash used here? , wouldn't a regular md5 or a faster hash suffice?

model/scikit/tests/test_scikit.py
unrahul: I would suggest used a  seed and generate it on the fly so that the source is clear, or do inline doc

model/scikit/dffml_model_scikit/scikit_base.py
0dust: [Transductive](https://scikit-learn.org/stable/glossary.html#term-transductive) methods (e.g. [OPTICS,](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.OPTICS.html#sklearn.cluster.OPTICS) [SpectralClustering](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.SpectralClustering.html#sklearn.cluster.SpectralClustering) ) can only make predictions on data which was used to train them. That's why for such methods the output corresponds to the predictions made on training data (calling `fit`  trains as well as predicts output as `labels_` for training data, also they don't have `predict` method ). This will probably have to be documented, informing the user about the methods which need to have prediction data same as training data.

tests/integration/test_models.py
0dust: Done.

model/scikit/tests/test_scikit.py
0dust: @pdxjohnny this is randomly generated, anyways I pushed the changes to generate data on the fly by changing seeds.

model/scikit/dffml_model_scikit/scikit_base.py
0dust: Ah! Everywhere in the `model` `sha384` has been used. I wrote it without even realizing that it might be a bit of overkill. But honestly, I am not sure. I guess md5 may suffice but using `sha384` here almost guarantees no collision.

0dust:
There are few nuances in `accuracy`. 
In clustering, we can score the model in [both](https://scikit-learn.org/stable/modules/classes.html#clustering-metrics) supervised and unsupervised way.
Supervised when we pass the ground truth and unsupervised when we don't.
Along with this, we have two different categories of methods [(transductive](https://scikit-learn.org/stable/glossary.html#term-transductive) and [inductive)](https://scikit-learn.org/stable/glossary.html#term-inductive) making a total of four possibilities for evaluating `accuracy` :-

1. Using transductive method when actual clusters(`TRUE_CLUSTER`) are known.

- Here I am expecting the user to add a feature `TRUE_CLUSTER` in the training data. This will be ignored in training but will be used to calculate the score by using `adjusted_mutual_info_score`. Because the method is transductive, predictions will be made on training data and there too `TRUE_CLUSTER` feature will be ignored. 

2. Using transductive method when actual clusters are not known.

- Simply pass the training data, model will make predictions on training data that will be used along with training data to calculate the score by `silhouette_score`.

3. Using inductive method when actual clusters are known.

- Similar to supervised learning. Add `TRUE_CLUSTER` feature in test data. The model will make predictions on test data and will use `TRUE_CLUSTER` to evaluate the prediction using `adjusted_mutual_info_score`.

4. Using inductive method when actual clusters are not known.

- Simply pass the test data, model will make predictions on test data that will be used along with test data to calculate the score by `silhouette_score`. This is similar to (2).

0dust:
@pdxjohnny @unrahul I am not sure how I should write good test cases for this, can you guide me through?

model/scikit/dffml_model_scikit/scikit_base.py
unrahul: yeah sha384 is an overkill to just to prevent collisions, also collisions here cannot be considered a show stopper or `threat, I would suggest using something like SHA1 if you have concerns with md5 and if you need added locality salt and hash. @pdxjohnny could comment better, there might be reasons for using SHA384 that I am not aware of.

model/scikit/dffml_model_scikit/scikit_base.py
```diff
@@ -119,6 +121,123 @@ def _filename(self):
...
+    async def predict(
+        self, repos: AsyncIterator[Repo]
+    ) -> AsyncIterator[Tuple[Repo, Any, float]]:
+        if not os.path.isfile(self._filename()):
+            raise ModelNotTrained("Train model before prediction.")
+        target = []
+        estimator_type = getattr(self.clf, "_estimator_type")
+        if estimator_type is "clusterer":
+            if "TRUE_CLUSTER" in set(self.features):
+                target.append("TRUE_CLUSTER")
```
pdxjohnny: Let's add another parameter to the config, called `tcluster` which defaults to `None`, if set, use the string given as the feature name which is the `TRUE_CLUSTER`

model/scikit/dffml_model_scikit/scikit_base.py
```diff
@@ -138,3 +257,25 @@ def _filename(self):
 
     async def __aexit__(self, exc_type, exc_value, traceback):
         Path(self._filename()).write_text(json.dumps(self.saved))
+
+
+class ScikitUnsprvised(Scikit):
+    def __init__(self, config) -> None:
+        super().__init__(config)
```
pdxjohnny: We can remove this `__init__`

model/scikit/dffml_model_scikit/scikit_base.py
unrahul: mm.. so SpetralClustering is O(n^3) and OPTICS is O(n^2) , I would say both implementations in its current form are irrelevant for practical analysis when the dataset is not *tiny*.

I would go with K-means and it that does not work by default on the data, some feature engineering (changing coordinate system for eg.) can help with clustering the data.. so is there a *real* need for adding this type of algorithms today?

model/scikit/dffml_model_scikit/scikit_base.py
```diff
@@ -119,6 +121,123 @@ def _filename(self):
             yield repo
 
 
+class ScikitContextUnsprvised(ScikitContext):
+    def __init__(self, parent):
+        super().__init__(parent)
+
+    def _feature_predict_hash(self):
```
pdxjohnny: Let's move this into `ScikitContext`

model/scikit/dffml_model_scikit/scikit_base.py
pdxjohnny: (This was the comment I meant to post, oops sorry) We do `sha384` this almost everywhere to avoid getting hit by bandit scans. Let's move to doing https://github.com/intel/dffml/issues/297 longterm

model/scikit/dffml_model_scikit/scikit_base.py
0dust: You are right. Both of these are computationally expensive and K-means remain unbeatable in most of the realistic scenarios. Considering this, I don't think there is any real need other than giving choice to the users to select what works best for their use case. So, should I drop them for now?

pdxjohnny: Lets warn with `self.logger.critical` when we hit this path. Mention that prediction must be done with the same data source that the model was trained with. Also give a warning about what Rahul said about needing to keep the dataset small.

pdxjohnny:
Merged as https://github.com/intel/dffml/commit/49137e7ae9bfd4c744f15540ca73a880a412a752 (because merging #315 caused a conflict with the changelog file I had to merge manually).

PR #294, opened 2020-01-07
source: idx1/3 source files 

sakshamarora1:


dffml/source/file.py
```diff
@@ -65,7 +65,7 @@ class FileSource(BaseSource):
         elif self.config.filename[::-1].startswith((".zip")[::-1]):
             opener = self.zip_opener_helper()
         else:
-            opener = open(self.config.filename, "r")
+            opener = open(self.config.filename, "rb")
```
pdxjohnny: Circle back on this change, and make sure all the other sources (json, csv) are still working with this change. We probably also need to change the various compression openers from `rt` to `rb`

pdxjohnny:
```python
opener = open(self.config.filename, self.READ_MODE)
```

then subclasses can override the READ_MODE and WRITE_MODE, we'll also need READ_MODE_COMPRESSED and WRITE_MODE_COMPRESSED

pdxjohnny: Create a new subclass BinaryFileSource with those properties set, then have IDX* source subclass from BinaryFileSource

docs/usage/mnist.rst
```diff
@@ -0,0 +1,28 @@
...
+=============
+
+This example will show you how to train a model on the MNIST dataset and use the
+model for prediction via the DFFML CLI and HTTP API.
+
+Download the files
+
+.. code-block:: console
+
+    $ curl -sSL 'http://yann.lecun.com/exdb/mnist/train-labels-idx1-ubyte.gz' \
```
pdxjohnny: This should be `idx3` (my bad, was going quick during the meeting)

docs/usage/mnist.rst
```diff
@@ -0,0 +1,92 @@
...
+``preprocess`` source what operations should be given to what features. And we
+need a way to specify what input gets what value. There will probably be some
+editing of the ``config_get`` code...
+
+This is just for quick and dirty pre-processing. Later we'll want to add a
+``dfpreprocess`` to use a dataflow to do pre-processing. For this example that
+would be too much overhead. Creating something that normalizes the dataset just
+by adding a few more CLI arguments will be very powerfull through it's
+simplicity.
+we need to figure out how to put a dataflow in front of
```
pdxjohnny: I think what we'll need to do is modify the config code in dffml/base py to be able to understand a typing.Dict value where the key is a str type, and then load the subsequent values based on their type. I'll try to explain more later.

pdxjohnny:
- For adding a section to tutorial on prediction. Complete the csv source modification and addition of configload for some image format that we can use
  - Need a picture of a handwritten digit in some file format
  - Need to write a ConfigLoader (see config/yaml) for whatever file extension you choose.
  - Make the configloader convert the file into the correct format (the array)
    - This means you’ll have to parse the file format, there might be a library out there to help you do this.

source/idx/dffml_source_idx/idx1.py
```diff
@@ -8,16 +10,17 @@
...
 
 
 @config
-class IDX1SourceConfig:
-    filename: str
-    feature: str
-    readwrite: bool = False
-    allowempty: bool = False
+class IDX1SourceConfig(BinaryFileSourceConfig):
+    feature: str = "unlabeled"
```
pdxjohnny: I think we should probably leave this without a default argument. Also, if you change `from dffml.base import config` to `from dffml.base import config, field` then you can add a help string for it (see model/tensorflow for examples).

source/idx/dffml_source_idx/idx3.py
```diff
@@ -10,16 +10,17 @@
...
 
 
 @config
-class IDX3SourceConfig:
-    filename: str
-    feature: str
-    readwrite: bool = False
-    allowempty: bool = False
+class IDX3SourceConfig(BinaryFileSourceConfig):
+    feature: str = "unlabeled"
```
pdxjohnny: Same thing with regards to removing the default

source/idx/dffml_source_idx/idx1.py
sakshamarora1: `feature` can't be left without a default argument as it gives this error because the config class in file.py have a parameter without default argument :-
``TypeError: non-default argument 'feature' follows default argument``
So a default value is necessary here.

pdxjohnny: Ahhh, that's because we're subclassing from `BinaryFileSourceConfig` we should make an issue to track this and make a fix for it eventually. For now could you just not subclass and just copy paste over the parameters from `BinaryFileSourceConfig`? Thanks! Then if we just cleanup the `docs/mnist.rst` file to only show the current training and accuracy commands, we can merge this and open a separate PR for the prediction stuff.

pdxjohnny:
Copying the MNIST handwritten digit database here due to network issues with the main site

Credits:
```
Yann LeCun, Courant Institute, NYU
Corinna Cortes, Google Labs, New York
Christopher J.C. Burges, Microsoft Research, Redmond
```

[t10k-labels-idx1-ubyte.gz](https://github.com/intel/dffml/files/4283895/t10k-labels-idx1-ubyte.gz)
[t10k-images-idx3-ubyte.gz](https://github.com/intel/dffml/files/4283896/t10k-images-idx3-ubyte.gz)
[train-labels-idx1-ubyte.gz](https://github.com/intel/dffml/files/4283897/train-labels-idx1-ubyte.gz)
[train-images-idx3-ubyte.gz](https://github.com/intel/dffml/files/4283898/train-images-idx3-ubyte.gz)

PR #331, opened 2020-01-18
WIP: Use auto args and config

jstar1:


dffml/df/memory.py
```diff
@@ -1065,7 +1072,8 @@ class MemoryOrchestratorConfig(BaseOrchestratorConfig):
     """
 
 
-class MemoryOrchestratorContextConfig(NamedTuple):
+@config
+class MemoryOrchestratorContextConfig:
```
pdxjohnny: We'll need to add

```python
@entrypoint("memory")
class MemoryOrchestrator(BaseOrchestrator, BaseMemoryDataFlowObject):
    CONFIG = MemoryOrchestratorConfig
    CONTEXT = MemoryOrchestratorContext
```

You'll also need to set the `CONFIG` property for all the non-`Context` classes. So anything that is a subclass of `BaseMemoryDataFlowObject`.

dffml/df/memory.py
```diff
@@ -205,7 +209,8 @@ def __call__(self) -> NotificationSetContext:
         return NotificationSetContext(self)
 
 
-class MemoryInputNetworkContextEntry(NamedTuple):
+@config
+class MemoryInputNetworkContextEntry:
```
pdxjohnny: Keep this as `NamedTuple`, let's only changed classes that end in `Config`

dffml/df/memory.py
jstar1: Just to clarify is this required for this merge request?

pdxjohnny: Ya, we'll need to remove the `args()` and `config()` methods as well, and this is what takes their place.

jstar1: Got it thanks @pdxjohnny

PR #333, opened 2020-01-20
repo: prediction: Predictions as feature specific dictionary.

### LINKED ISSUE #106: repo: prediction: Should be feature specific
Right now `Repo` can only hold one prediction at a time. This doesn't make sense as lots of things could be the result of a prediction. Prediction should be a key value mapping:

```json
{
    "name_of_predicted_value": {
        "value": "outcome_of_Model.predict",
        "confidence": 0.42
    },
    "meaning_of_life": {
        "value": 42,
        "confidence": 1.0
    }
}
```

We'll want to make `prediction` similar to `features` in that it will be a `dict`.

https://github.com/intel/dffml/blob/2965d02040f5ef3c04e108af27a2afa369b21614/dffml/repo.py#L62-L63

We'll want to add a new argument to the `predicted` method, which will be the new first argument, which will be the name of the feature we are predicting. This will be the key in the `prediction` `dict` for the value and confidence given. We need to to this by creating a `RepoPrediction` instance and using that as the value in the `dict`.

https://github.com/intel/dffml/blob/2965d02040f5ef3c04e108af27a2afa369b21614/dffml/repo.py#L209-L215

We need to change the `prediction` method of `Repo` to take the name of the feature we want the `RepoPrediction` of. We'll use the feature name as the key for the `prediction` `dict` and return the value, which will be of type `RepoPrediction`.

https://github.com/intel/dffml/blob/2965d02040f5ef3c04e108af27a2afa369b21614/dffml/repo.py#L218-L222

We'll also need to change all the `predict` methods in every model so that they pass the feature name they are predicting (this is `self.config.parent.predict` for scikit, tensorflow, and scratch).

The testing model for CLI tests will also need to be changed:

https://github.com/intel/dffml/blob/2965d02040f5ef3c04e108af27a2afa369b21614/tests/test_cli.py#L127-L130

aghinsa:
fixes #106


aghinsa:
I havn't changed scikit models yet,also what should we do with [dumping csv](https://github.com/intel/dffml/blob/050833e15c73aa30f4b85ced059e5fb4dacdd384/dffml/source/csv.py#L196) as in what format.

dffml/operation/model.py
```diff
@@ -36,10 +36,10 @@ def __post_init__(self):
 async def model_predict(self, features: Dict[str, Any]) -> Dict[str, Any]:
     async def repos():
         yield Repo("", data={"features": features})
-
+    target_name = self.config.model.config.predict.NAME
     async for repo in self.mctx.predict(repos()):
         return {
             "prediction": {
```
pdxjohnny: This should just be
```python
return {
    "prediction": { repo.prediction() }
}
```

pdxjohnny: since `repo.prediction()` will now have the feature names as the first level keys in their `dict`

dffml/repo.py
```diff
@@ -206,17 +207,17 @@ def feature(self, name: str) -> Any:
...
         Set the prediction for this repo
         """
-        self.data.prediction = RepoPrediction(
+        self.data.prediction[target] = RepoPrediction(
             value=value, confidence=float(confidence)
         )
         self.data.last_updated = datetime.now()
 
-    def prediction(self) -> RepoPrediction:
+    def prediction(self,target:str) -> RepoPrediction:
```
pdxjohnny: Let's copy the way `feature()/features()` works. So let's make two methods.

dffml/source/csv.py
```diff
@@ -113,7 +114,18 @@ class CSVSource(FileSource, MemorySource):
             repo_data = {}
             # Parse headers we as the CSV source added
             csv_meta = {}
-            for header in self.CSV_HEADERS:
+            row_keys=[]
+            # getting all keys starting with "prediction","confidence"
+            for _header in self.CSV_HEADERS:
+                row_keys.extend(list(
+                     filter(
+                    lambda x: x.startswith(_header),
```
pdxjohnny: I think we want `_header + "_"` here. Also, variables shouldn't be prefixed with `_` unless we aren't using them (reference: https://stackoverflow.com/a/5893946/12310488 bullet point 3)

model/scikit/tests/test_scikit.py
```diff
@@ -129,9 +129,11 @@ def tearDownClass(cls):
 
     async def test_02_predict(self):
         async with self.sources as sources, self.model as model:
+            target = model.config.predict.NAME
             async with sources() as sctx, model() as mctx:
                 async for repo in mctx.predict(sctx.repos()):
-                    prediction = repo.prediction().value
+                    # prediction = repo.prediction(target).value
+                    prediction = repo.prediction(target)["value"]
```
pdxjohnny: This shouldn't have changed to `['value']` we need to make sure we're creating `RepoPrediction` instances.

model/tensorflow/dffml_model_tensorflow/dnnc.py
```diff
@@ -348,13 +345,13 @@ class DNNClassifierModel(Model):
...
         $ head iris_training.csv
-        $ sed -i 's/.*setosa,versicolor,virginica/SepalLength,SepalWidth,PetalLength,PetalWidth,classification/g' *.csv
+        $ sed -i 's/.*setosa,versicolor,virginica/SepalLength,SepalWidth,PetalLength,PetalWidth,predict/g' *.csv
         $ head iris_training.csv
         $ dffml train \\
             -model tfdnnc \\
             -model-epochs 3000 \\
             -model-steps 20000 \\
-            -model-classification classification:int:1 \\
+            -model-predict predict:int:1 \\
```
pdxjohnny: Because we want to `-predict` the `classification`

model/tensorflow/dffml_model_tensorflow/dnnc.py
```diff
@@ -139,10 +139,10 @@ def model(self):
 
 @config
 class DNNClassifierModelConfig:
-    classification: Feature = field(
-        "Feature name holding classification value"
+    predict: Feature = field(
```
pdxjohnny: We should really do the `classification` -> `predict` in a seperate pull request. It's currentl in both `sql_op` and this. And it's really it's own thing.

dffml/operation/model.py
aghinsa: did you mean `repo.predictions()` after we adopt `feature/features` style

pdxjohnny: Yes, sorry. I had commented this comment before that comment. You are correct.

aghinsa: Okay.Thanks

model/tensorflow/dffml_model_tensorflow/dnnc.py
aghinsa: Yeah,sorry my bad,created [PR](https://github.com/intel/dffml/pull/340)

pdxjohnny: Merged, thanks!

.ci/run.sh
```diff
@@ -75,7 +75,6 @@ function run_plugin() {
   fi
 
   if [ "x${GITHUB_ACTIONS}" == "xtrue" ] && [ "x${GITHUB_REF}" == "xrefs/heads/master" ]; then
-    git status
```
pdxjohnny: Did you fetch origin before you did the diff? That was an important part...

pdxjohnny: You might want to do `git log --walk-reflogs repo_predict` and checkout the one before the reset to master

Issue #233, opened 2019-10-25
repo: Repo.src_url should be Repo.key


pdxjohnny:
DFFML started as a project centred around Git repos. As such, the unique key for a Git repo was the upstream source URL. `src_url` was used as the parameter to hold the unique key.

Now that DFFML has become more general purpose, we need to change *all* references of `src_url` to `key`. Hopefully this is as simple as find and replace...

0dust:
@pdxjohnny _all_ references of `src_url` means literally **all** references in the entire code, or only references of `src_url` in `repo.src_url` ?

pdxjohnny:
@0dust **all** meaning the entire codebase

Issue #303, opened 2020-01-09
feature: Replace Feature with DefFeature


pdxjohnny:
We need to move from class based approach for `Feature`s to instance based.

Right now features are still setup for evaluation (#289) this makes no sense with the rest of the codebase at this point. We must clean this up.

Other than that we also need to make `Feature` a dataclass or namedtuple or something. The point here being that `.name()` should just be `.name` and `.NAME` shouldn't exist anymore (same for dtype).

Naeemkh:
@pdxjohnny, I am working on this issue.

pdxjohnny:
@Naeemkh Awesome! This will be great!

pdxjohnny:
Essentially, we want to remove this code:

https://github.com/intel/dffml/blob/4c1e7870dba12c298d5a7f4740a23af634cacb7e/dffml/feature/feature.py#L160-L187

And replace

https://github.com/intel/dffml/blob/4c1e7870dba12c298d5a7f4740a23af634cacb7e/dffml/feature/feature.py#L19

With

```python
@dataclasses.dataclass
class Feature:
    name: str
    dtype: Type # from typing import Type
    length: Union[int, List[int]]

    # ... All the existing methods in feature (but not dtype and length)
```

We'll also go through and remove all references to `DefFeature` and replace them with `Feature`

pdxjohnny:
Right now to create a new `Feature` you have to make a whole new `class`. What we're trying to do here is have `Feature` be a class. And have features be instances of that class. Rather than other classes which are subclasses of `Feature`.

Issue #289, opened 2020-01-06
feature: Remove evaluation code


pdxjohnny:
There is a bunch of code in `dffml/feature/feature.py` that isn't being used. For example the `calc`, `parse`, and `fetch` methods. And all the stuff relating to running those methods from the `Features` class.

Yash-Varshney:
thanks for the issue @pdxjohnny. I will like to work on it, I have a question do we just remove those methods or should we comment them out ( if they are needed in coming future ) ?

pdxjohnny:
Hi @Yash-Varshney, they should be removed. You'll need to go through the test code to and remove the tests associated with them. Thanks!

Yash-Varshney:
@pdxjohnny  please review the commit !!

pdxjohnny:
@Yash-Varshney I'll get to this soon. Sorry I'm pretty busy today. Hopefully today or tomorrow. Can I get you do take these changes and apply them on top of the master branch? I think you'll just need to

```console
$ git fetch --all
$ git merge origin/master
```

and then push.

Thanks!

Yash-Varshney:
@pdxjohnny  I did that and it says that everything is up-to-date.

PR #354, opened 2020-01-27
repo: src_url to key

0dust:
Issue #233


dffml/feature/feature.py
```diff
@@ -75,9 +75,9 @@ class Data(Task):
 
     LOGGER = LOGGER.getChild("Data")
 
-    def __init__(self, src_url: str) -> None:
-        super().__init__(key=src_url)
-        self.src_url = src_url
+    def __init__(self, key: str) -> None:
+        super().__init__(_key=key)
+        self.key = key
```
0dust: Probably name should not be `_key`.

dffml/source/csv.py
```diff
@@ -121,12 +121,12 @@ class CSVSource(FileSource, MemorySource):
                     del row[header]
             # Set the features
             features = {}
-            for key, value in row.items():
-                if value != "":
```
0dust: `key` renamed to `_key` because we are already using `key` which gets overwritten. For consistency `value` changed to `_value`

dffml/util/monitor.py
```diff
@@ -37,21 +37,21 @@ class Task(object):
 
     LOGGER = LOGGER.getChild("Task")
 
-    def __init__(self, func=None, key: Any = "") -> None:
+    def __init__(self, func=None, _key: Any = "") -> None:
         coro = None
```
0dust: `key` changed to `_key`, this is related to `feature.py`.

dffml/feature/feature.py
pdxjohnny: This doesn't really matter, it'll be removed via #303 and #289

PR #357, opened 2020-01-27
types : definition : Add validate to definition

### LINKED ISSUE #349: df: types: Definition validation
Attach a `validate` property to `Definition` which will get run upon instantiation of an input, similar to #86 

The validate property will be a callable (function or lambda) which returns the sanitized version of the `value` (which is then set to the `Input.value`).

aghinsa:
fixes #349


dffml/df/types.py
```diff
@@ -37,6 +37,9 @@ class Definition(NamedTuple):
     lock: bool = False
     # spec is a NamedTuple which could be populated via a dict
     spec: NamedTuple = None
+    # validate property will be a callable (function or lambda) which returns
+    # the sanitized version of the value
+    validate : Callable[["primitive"],"primitive"] = None
```
pdxjohnny: I think we want `typing.Any` here instead of `"primitive"`.

tests/integration/test_service_dev.py
```diff
@@ -2,9 +2,11 @@
 This file contains integration tests. We use the CLI to exercise functionality of
 various DFFML classes and constructs.
 """
+import io
+import json
 import inspect
 import pathlib
-
+import unittest
```
pdxjohnny: I accidentally removed these when I went through and removed all unused import. The CI wasn't failing... weird. I figured out what was wrong: https://github.com/intel/dffml/issues/358

If you merge master into this branch we'll be good to go.

tests/test_types.py
```diff
@@ -0,0 +1,60 @@
...
+        "shape": {
+            "name": name,
+            "radius": radius,
+            "area": pie * radius * radius,
+        }
+    }
+
+
+class TestDefintion(AsyncTestCase):
+    async def test_validate(self):
```
pdxjohnny: Let's add another `test_` where the `pie` datatype raises a `InputValidationError` (add this new error class to `dffml/exceptions.py`).

Use `with self.assertRaises(InputValidationError)` in the new test to make sure that we catch it.

PR #387, opened 2020-02-17
Input forwarding to subflows

### LINKED ISSUE #368: df: Input forwarding to subflows
Within DataFlow, allow for specifying input definitions / origins that should be forwarded directly into a subflow.

`run_dataflow` takes the same arguments that `OrchestratorContext.run` does. It would be great if the parent flow knew about which subflows were active within it, and could forward `Input`s that enter it's network to any given subflow (probably based off the `instance_name`).

### LINKED ISSUE #386: df: memory: Insure operation instance name gets set upon instantiation
https://github.com/intel/dffml/blob/66116d914d1459c38dd77de165a349b35cffa6a7/dffml/df/memory.py#L844-L846

We need to make sure the operation implementation get the operation set to one with the correct `instance_name`, ( I think the `op._replace` method is needed, not sure).

Reporter: @aghinsa

aghinsa:
fixes #386 
fixes #368


tests/df/test_forward.py
```diff
@@ -0,0 +1,124 @@
...
+                    value=[VECTOR.name],
+                    definition=GetSingle.op.inputs["spec"],
+                ),
+                # Input(value=[2, 3], definition=SHAPE),
+                # Input(
+                #     value=[collect_data.op.outputs["data"].name],
+                #     definition=GetSingle.op.inputs["spec"],
+                # ),
+            ],
+            implementations={
```
aghinsa: If i ran it like this,
```
Debug ctx_str:feature1,results:{'flow_results': {'feature1': {'vector': [1, 2, 3, 1, 2, 3]}}}
```
This is the result,doesn't that mean its getting forwarded,but if i uncomment the other two,it rises
```
dffml.df.exceptions.DefinitionNotInContext: feature1: mapping
```
whats going wrong here?

dffml/df/memory.py
```diff
@@ -841,6 +841,7 @@ def __init__(
                 opimp = OperationImplementation.load(operation.name)
             else:
                 raise OperationImplementationNotInstantiable(operation.name)
+        opimp.op = opimp.op._replace(instance_name=operation.instance_name)
```
pdxjohnny: Nice

tests/operation/test_dataflow.py
```diff
@@ -2,11 +2,12 @@
 import json
 import unittest.mock
 
-from dffml.df.types import DataFlow, Input
+from dffml.df.types import DataFlow, Input, Definition
```
pdxjohnny: Let's merge master, I think these changes are showing up because of the unused imports being removed.

dffml/df/memory.py
```diff
@@ -1207,6 +1230,25 @@ def __init__(
...
 
+    async def forward_inputs_to_subflow(self, inputs: List):
+        # Go through input set,find instance_names of registered subflows which
+        # have definition of the current input listed in `forward`.
+        # If found,add `input` to list of inputs to forward for that instance_name
+        forward = self.config.dataflow.forward
+        for ip in inputs:
+            instance_list = forward.get_instances_to_forward(ip.definition)
+            for instance_name in instance_list:
+                self.inputs_to_forward.setdefault(instance_name, []).append(ip)
```
pdxjohnny: `setdefault(...).append()` - fucking beautiful, very nice.

dffml/df/memory.py
```diff
@@ -1207,6 +1239,33 @@ def __init__(
...
 
+    async def forward_inputs_to_subflow(self, inputs: List[Input]):
+        # Go through input set,find instance_names of registered subflows which
+        # have definition of the current input listed in `forward`.
+        # If found,add `input` to list of inputs to forward for that instance_name
+        forward = self.config.dataflow.forward
+        for item in inputs:
+            instance_list = forward.get_instances_to_forward(item.definition)
+            for instance_name in instance_list:
+                self.inputs_to_forward.setdefault(instance_name, []).append(
```
pdxjohnny: I love `setdefault().append(...)`, beautiful.

dffml/operation/dataflow.py
```diff
@@ -33,8 +54,10 @@ class RunDataFlowConfig:
             )
             for val_def in val_defs
         ]
-
     async with self.octx.parent(self.config.dataflow) as octx:
+        # Register the subflow with parent,so that parent flow can
+        # forward inputs of specifed defintions(in DataFlow.forward)
+        await self.octx.register_subflow(self.parent.op.instance_name, octx)
```
pdxjohnny: Let's either now, or in the future, make this into one call, so instead of:

```python

    async with self.octx.parent(self.config.dataflow) as octx:
        # Register the subflow with parent,so that parent flow can
        # forward inputs of specifed defintions(in DataFlow.forward)
        await self.octx.register_subflow(self.parent.op.instance_name, octx)
```

It would be:

```python
    async with self.subflow(self.config.dataflow) as octx:
```

So `subflow` would need to be a method of `OperationImplementationContext`

aghinsa: Yes!

aghinsa: wait ,i'm confused,`OperationImplementationContext` or `OperationImplementationNetworkContext` ?

pdxjohnny: In `OperationImplementationContext`

pdxjohnny: You'll need to use [contextlib.asynccontextmanager](https://docs.python.org/3/library/contextlib.html?highlight=contextlib#contextlib.asynccontextmanager) because you need to `yield` within an `async def` function

pdxjohnny:
- Let not have a place in memory where inputs get appended and then added to subflows when they start new contexts (for now, we may reassess this later)
- Lets only forward inputs to active contexts within a subflow (which was in a previous revision of this PR)

pdxjohnny:
@aghinsa I had a realization on this, let me know your thoughts and if you want to take a stab at it: https://github.com/intel/dffml/issues/415

dffml/df/base.py
```diff
@@ -27,6 +27,8 @@
 from ..util.asynchelper import context_stacker
 from ..util.entrypoint import base_entry_point
 
+from contextlib import asynccontextmanager
```
pdxjohnny: Can we move this up with the rest of the imports from the standard library

tests/operation/test_dataflow.py
```diff
@@ -3,6 +3,7 @@
 from dffml.operation.dataflow import run_dataflow, RunDataFlowConfig
 from dffml.operation.output import GetSingle
 from dffml.util.asynctestcase import AsyncTestCase
+from dffml.df.base import op
```
pdxjohnny: Did you mean to change this file?

dffml/df/memory.py
```diff
@@ -1202,6 +1219,26 @@ def __init__(
...
+                    item
+                )
+        self.logger.debug(
+            f"Forwarding inputs from {self.inputs_to_forward} to {self.subflows}"
+        )
+        for instance_name, inputs in self.inputs_to_forward.items():
```
pdxjohnny: It looks like `inputs_to_forward` could be made local to this function?

dffml/df/types.py
```diff
@@ -122,6 +122,7 @@ class Operation(NamedTuple, Entrypoint):
     conditions: Optional[List[Definition]] = []
     expand: Optional[List[str]] = []
     instance_name: Optional[str] = None
+    register_with_parent_flow: bool = False
```
pdxjohnny: Do we need this?

dffml/df/base.py
```diff
@@ -86,6 +88,18 @@ def config(self):
         object associated with this operation implementation context.
         """
 
+    @asynccontextmanager
+    async def subflow(self, dataflow):
+        """
+        Registers subflow `dataflow` with parent flow and returns an instance of `BaseOrchestratorContext`
```
pdxjohnny:
```suggestion
        Registers subflow `dataflow` with parent flow and yields an instance of `BaseOrchestratorContext`
```

dffml/df/base.py
```diff
@@ -86,6 +88,18 @@ def config(self):
...
+
+        >>> async with self.subflow(self.config.dataflow) as octx:
+            octx.run(test_inputs)
```
pdxjohnny: This is going to cause a doctest error (and will fail the CI once #431 is fixed) 

How about:

```python
        >>> async def my_operation(arg):
        ...     async with self.subflow(self.config.dataflow) as octx:
        ...         return octx.run({"ctx_str": []})
```

dffml/df/base.py
```diff
@@ -86,6 +87,19 @@ def config(self):
         object associated with this operation implementation context.
         """
 
+    @asynccontextmanager
+    async def subflow(self, dataflow):
+        """
+        Registers subflow `dataflow` with parent flow and yields an instance of `BaseOrchestratorContext`
+
+        >>> async def my_operation(arg):
+                async with self.subflow(self.config.dataflow) as octx:
```
pdxjohnny:
```suggestion
        ...     async with self.subflow(self.config.dataflow) as octx:
```

dffml/df/base.py
```diff
@@ -86,6 +87,19 @@ def config(self):
...
+                    return octx.run({"ctx_str": []})
```
pdxjohnny:
```suggestion
        ...         return octx.run({"ctx_str": []})
```

dffml/df/memory.py
```diff
@@ -1072,6 +1085,9 @@ def __init__(
     ) -> None:
         super().__init__(config, parent)
         self._stack = None
+        # Maps instance_name to OrchestratorContext
+        self.subflows = {}
+        # Maps instance_name to List[Input]
```
pdxjohnny:
```suggestion
```

CHANGELOG.md
```diff
@@ -289,26 +341,34 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
 - Corrected maxsplit in util.cli.parser
 - Check that dtype is a class in Tensorlfow DNN
 - CI script no longer always exits 0 for plugin tests
 - Corrected render type in setup.py to markdown
 
 ## [0.1.2] - 2019-03-29
+
 ### Added
+
```
pdxjohnny: I think this happened to someone else recently because of their editors config, but let's remove these newlines.

dffml/df/types.py
```diff
@@ -361,6 +361,29 @@ def _fromdict(cls, **kwargs):
...
+        self._internal_book.extend(definition_list)
+
+    def get_instances_to_forward(self, definition):
+        if not definition in self._internal_book:
+            return []
+        return [
+            instance_name
+            for instance_name, definitions in self.book.items()
+            if definition in definitions
+        ]
```
pdxjohnny: `export` and `_fromdict` in `DataFlow` need to be updated to correctly import and export this data structure.

dffml/df/base.py
pdxjohnny: The `...` is needed for the doctests.

dffml/df/memory.py
pdxjohnny: I think this comment was associated with the `dict` we moved into the other function.

CHANGELOG.md
aghinsa: Ughh!! I had auto-format on.

pdxjohnny: :( sorry!

pdxjohnny:
Nice!! Thanks man! :D

PR #404, opened 2020-02-21
Fix spelling errors

us:


us:
Lint give error about changelog. But it's so small change so I didn't need to add

PR #407, opened 2020-02-25
WIP: Python API example usage for scikit model

### LINKED ISSUE #384: docs: model: scikit: Include examples of how to use from Python API
The model plugins page has command line examples but needs examples from Python as well.

Copy the format of the `examples/quickstart_filenames.py` file and add a `.. code-block:: python` section for each model, keeping the Python example consistent with the command line example for each model. Let's add the python example after the CLI example for each.

For scikit, `model/scikit/dffml_model_scikit/__init__.py` is what needs to be update.

For scratch and tensoflow, the docstrings of the `Model`s themselves will need to be updated.

sudharsana-kjl:
fix #384


pdxjohnny:
I think that the usage of `requests` might be better suited under a page like the javascript one (just added in #418), but for Python, under the HTTP API docs. Issue #384 was meant to be about modifying the quickstart.py code to use respective models within each model docstring. This is great for a Python HTTP API example, could you make this similar to the `service/http/docs/javascript.rst` but as `python.rst`?

sudharsana-kjl:
@pdxjohnny I've modified `model/scikit/dffml_model_scikit/__init__.py` and added examples for LinearRegression and Kmeans. I moved the example using requests to python.rst as suggested. Could you please review it?

PR #441, opened 2020-03-01
Update scikit_models.py

### LINKED ISSUE #429: model: scikit: Change help and default for predict config of unsupervised models
`Prediction` should be `cluster` since it will be used as the name for what the output of the model is. For example, a prediction currently results in:

```json
[
    {
        "extra": {},
        "features": {
            "Col1": 6.09809669,
            "Col2": 8.36434181,
            "Col3": 6.70940915,
            "Col4": -7.91491768
        },
        "key": "0",
        "last_updated": "2020-02-29T07:26:20Z",
        "prediction": {
            "Prediction": {
                "confidence": 0.6365141682948129,
                "value": 7
            }
        }
    }
]
```

If would be more clear if it said `cluster` in place of `Prediction`.

https://github.com/intel/dffml/blob/3791f76d802fe6f3a63f9991c8b5b4afde6f7c76/model/scikit/dffml_model_scikit/scikit_models.py#L240-L247

Also, the help text says it's there for compatibility reasons. It should say something like `"Name used as meaning of prediction"`

iamandeepsandhu:
Fixes #429

### DIFF
--- a/model/scikit/dffml_model_scikit/__init__.py
+++ b/model/scikit/dffml_model_scikit/__init__.py
@@ -320,7 +320,7 @@ Example usage of KMeans Clustering Model using python API:
         model,
         {"Col1": 6.09809669, "Col2": 8.36434181, "Col3": 6.70940915, "Col4": -7.91491768},
     ):
-        features["cluster"] = prediction["Prediction"]["value"]
+        features["cluster"] = prediction["cluster"]["value"]
         print(features)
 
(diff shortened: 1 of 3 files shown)


pdxjohnny:
Sweet! You'll need to update the tests in `model/scikit/tests` as well for this to work. Also, this line needs to be updated:

https://github.com/intel/dffml/blob/aad53e902dc8ce7b9d21b949ac8d1a960969b280/model/scikit/dffml_model_scikit/__init__.py#L323

Issue #402, opened 2020-02-21
source: db: Source using the database abstraction


pdxjohnny:
In https://github.com/intel/dffml/commit/7a80ac8bea7af104aea2290b6761737f594bb320 the database abstraction was added. It would be nice if we could leverage that to create a source which takes a database in it's config. So that whatever database abstractions people implement (such as [sqlite](https://github.com/intel/dffml/commit/7a80ac8bea7af104aea2290b6761737f594bb320) and [mysql](https://github.com/intel/dffml/commit/29eeb2e7136fc8e1659c63e4b9ab88963384cd3d)) this source can use any of them.

Lactem:
@pdxjohnny would you please let me know if [my skeleton](https://github.com/Lactem/dffml/commit/14efc0e2286b01578a9e3e21c2d3e00ecdde7ab0) is on the right path? I realize I didn't use the dffml service command to auto-copy the whole skeleton, but I'll get to that eventually. It'll help me understand the project better if I copy over each file manually.

Were you thinking the source would have a config option that says which db to point to, and then the source would use the db's config and abstracted insert/update/remove/etc methods to access records? I want to make sure I'm on the right track before I fill in too much.

pdxjohnny:
> @pdxjohnny would you please let me know if [my skeleton](https://github.com/Lactem/dffml/commit/14efc0e2286b01578a9e3e21c2d3e00ecdde7ab0) is on the right path? I realize I didn't use the dffml service command to auto-copy the whole skeleton, but I'll get to that eventually. It'll help me understand the project better if I copy over each file manually.

Awesome! It's very close, I left a few comments.

> Were you thinking the source would have a config option that says which db to point to, and then the source would use the db's config and abstracted insert/update/remove/etc methods to access records? I want to make sure I'm on the right track before I fill in too much.

I left some comments about the config, let me know if anything is unclear. I'd say try making a source that doesn't do anything (`return` right away for each of the required methods), figure out how to get the test_dbsource.py set up. Then use the abstracted insert/update/remove/etc methods to access records. Not sure if this helps, let me know when you're ready for more feedback.

Thanks! :)

Lactem:
Thanks for all the guidance. I haven't had a chance to look at your feedback yet, but it looks like I have a lot to work with. I'll plan to push what I have later this week.

pdxjohnny:
[...]

PR #456, opened 2020-03-06
source: db: Source using the database abstraction

Lactem:
Closes issue #402


source/db/dffml_source_db/source.py
```diff
@@ -0,0 +1,143 @@
+import collections
```
pdxjohnny: I can't remember if I said to make this db plugin or not. So sorry if I did, but either way, we're going to go ahead and not have this be a plugin. Since it has no external dependencies.

So this `source/db/dffml_source_db/source.py` file should really go in `dffml/source/db.py`

source/db/dffml_source_db/source.py
```diff
@@ -0,0 +1,143 @@
...
+from dffml.db.sqlite import SqliteDatabase
+from dffml.record import Record
+from dffml.source.source import BaseSource, BaseSourceContext
+from dffml.util.entrypoint import entrypoint
+
+
+@config
+class DbSourceConfig(BaseConfig):
+    db_implementation: str
+    db_config: any
```
pdxjohnny: `@config` should take care of all the loading for us. So we can just put:

```suggestion
    db: BaseDatabase
```

source/db/dffml_source_db/source.py
```diff
@@ -0,0 +1,143 @@
...
+from dffml.db.base import BaseDatabase, Condition
+from dffml.db.sqlite import SqliteDatabase
+from dffml.record import Record
+from dffml.source.source import BaseSource, BaseSourceContext
+from dffml.util.entrypoint import entrypoint
+
+
+@config
+class DbSourceConfig(BaseConfig):
+    db_implementation: str
```
pdxjohnny:
```suggestion
```

source/db/tests/test_source.py
```diff
@@ -0,0 +1,81 @@
...
+        file, cls.database_name = tempfile.mkstemp(suffix=".db")
+        os.close(file)
+
+        # Sqlite config
+        cls.db_config = SqliteDatabaseConfig(cls.database_name)
+
+        # DbSource config
+        cls.source_config = DbSourceConfig(
+            db_implementation="sqlite",
+            db_config=cls.db_config,
```
pdxjohnny:
```suggestion
            db=SqliteDatabase(cls.db_config),
```

source/db/tests/test_source.py
```diff
@@ -0,0 +1,81 @@
...
+        # Sqlite db file
+        file, cls.database_name = tempfile.mkstemp(suffix=".db")
+        os.close(file)
+
+        # Sqlite config
+        cls.db_config = SqliteDatabaseConfig(cls.database_name)
+
+        # DbSource config
+        cls.source_config = DbSourceConfig(
+            db_implementation="sqlite",
```
pdxjohnny:
```suggestion
```

source/mysql/tests/test_source.py
```diff
@@ -67,7 +67,6 @@ def setUpClass(cls):
     @classmethod
     def tearDownClass(cls):
         super().tearDownClass()
-        cls._exit_stack.__exit__(None, None, None)
```
pdxjohnny: Was there a reason this was removed?

source/db/dffml_source_db/source.py
```diff
@@ -0,0 +1,143 @@
...
+                    ]
+                else:
+                    key_value_pairs[key] = 1
+            else:
+                key_value_pairs[key] = record.data.__dict__[key]
+        async with self.parent.db_impl() as db_ctx:
+            # Insert AND update since some tests don't insert before updating
+            # TODO: Maybe change the sqlite update query to use INSERT INTO ... ON DUPLICATE KEY UPDATE
+            # Note that this would impose all other db implementations to use "update or create if not exists" functionality
+            await db_ctx.insert(self.parent.config.table_name, key_value_pairs)
```
pdxjohnny: Will this work?

https://github.com/intel/dffml/blob/cf4ee851414bef9dab33b5c037afefd6577932fa/dffml/db/sqlite.py#L73-L88

source/db/dffml_source_db/source.py
```diff
@@ -0,0 +1,143 @@
...
+            self.db_impl = SqliteDatabase(self.config.db_config)
+        # TODO: Add more implementations. Is there a better way to do this?
+        else:
+            raise RuntimeError(
+                f"No implementation named {self.config.db_implementation} found for DbSource abstraction"
+            )
+        return self
+
+    async def __aexit__(self, exc_type, exc_value, traceback):
+        await self.db_impl.__aexit__(exc_type, exc_value, traceback)
```
pdxjohnny: It looks like we've got an `__aexit__` without an `__aenter__`.

source/db/dffml_source_db/source.py
```diff
@@ -0,0 +1,143 @@
...
+    async def __aenter__(self) -> "DbSource":
+        # Instantiate a subclass of BaseDatabase using "db_implementation" from DbSourceConfig
+        self.db_impl = None
+        if self.config.db_implementation == "sqlite":
+            self.db_impl = SqliteDatabase(self.config.db_config)
+        # TODO: Add more implementations. Is there a better way to do this?
+        else:
+            raise RuntimeError(
+                f"No implementation named {self.config.db_implementation} found for DbSource abstraction"
+            )
```
pdxjohnny: If we switch to the single `db: BaseDatabase` in the config I think we can remove all this and replace it with:

```python
    async def __aenter__(self) -> "DbSource":
        self.db_impl = await self.config.db.__aenter__()
```

source/db/dffml_source_db/source.py
```diff
@@ -0,0 +1,143 @@
...
+
+@entrypoint("dbsource")
+class DbSource(BaseSource):
+    CONFIG = DbSourceConfig
+    CONTEXT = DbSourceContext
+    db_impl: BaseDatabase
+
+    def __init__(self, cfg: Type[BaseConfig]) -> None:
+        super().__init__(cfg)
+
```
pdxjohnny:
```suggestion
        self.db_impl = None

```

source/db/dffml_source_db/source.py
```diff
@@ -0,0 +1,143 @@
...
+                )
+            )
+        return record
+
+
+@entrypoint("dbsource")
+class DbSource(BaseSource):
+    CONFIG = DbSourceConfig
+    CONTEXT = DbSourceContext
+    db_impl: BaseDatabase
```
pdxjohnny: This should be set at the instance level rather than the class level (so in `__init__`) . Also we could probably have it just be called `self.db`.

source/mysql/tests/test_source.py
Lactem: It would throw this error: `AttributeError: type object 'TestDbSource' has no attribute '_exit_stack'`

I'm not too sure how the _exit_stack stuff works, so I didn't see the use it added.

Lactem:
@pdxjohnny Not sure why the DOCS test is failing now. Running ./scripts/docs.sh doesn't produce any new files on my end. Also, should I squash these commits before removing the WIP?

Btw, is there a travis command or something to run all of these tests locally? `python3.7 setup.py test` doesn't hit everything.

pdxjohnny:
> @pdxjohnny Not sure why the DOCS test is failing now. Running ./scripts/docs.sh doesn't produce any new files on my end.

It's an annoying thing we recently fixed (I believe), if you merge master it should go away

> Also, should I squash these commits before removing the WIP?

I'll squash when I merge, no worries on that

> Btw, is there a travis command or something to run all of these tests locally? `python3.7 setup.py test` doesn't hit everything.

Unfortunately this is an area we need to improve on. I've opened https://github.com/intel/dffml/issues/490 to track

source/mysql/tests/test_source.py
pdxjohnny: This is in mysql source though?

tests/source/test_db.py
```diff
@@ -0,0 +1,80 @@
...
+        os.close(file)
+
+        # Sqlite config
+        cls.db_config = SqliteDatabaseConfig(cls.database_name)
+
+        # DbSource config
+        cls.source_config = DbSourceConfig(
+            db=SqliteDatabase(cls.db_config),
+            table_name=cls.table_name,
+            model_columns="key feature_PetalLength feature_PetalWidth feature_SepalLength feature_SepalWidth target_name_confidence target_name_value",
```
pdxjohnny: I looks like you hae 
```suggestion
            model_columns="key feature_PetalLength feature_PetalWidth feature_SepalLength feature_SepalWidth target_name_confidence target_name_value".split(),
```

It looks like you have model_columns defined as `List[str]` in the config

Lactem:
@pdxjohnny Tests now pass :D thanks for all the help! Are there other changes I should make before merging?

Lactem:
> Sweet!! Just one thing

Definitely. Not sure why I removed it.

PR #483, opened 2020-03-21
Memory:dataflow:Cleanup in memory

### LINKED ISSUE #452: df: memory: Move operations_parameter_set_pairs to MemoryOrchestratorContext
`operations_parameter_set_pairs` really has no reason to be in `MemoryOperationImplementationNetworkContext`. It should really live in `MemoryOrchestratorContext`.

### LINKED ISSUE #477: df: memory: Change MemoryInputSet unvalidated removal
`remove_unvalidated_inputs` and `remove_input` should really be `async` functions. Also, they should be required method defined within the abstract base class.

aghinsa:
fixes #477 
fixes #452

### DIFF
--- a/dffml/df/memory.py
+++ b/dffml/df/memory.py
@@ -1559,9 +1559,7 @@ class MemoryOrchestratorContext(BaseOrchestratorContext):
         # Identify which operations have complete contextually appropriate
         # input sets which haven't been run yet and are stage operations
         async for operation, parameter_set in self.operations_parameter_set_pairs(
-            ctx,
-            self.config.dataflow,
-            stage=stage,
+            ctx, self.config.dataflow, stage=stage
         ):
             # Add inputs and operation to redundancy checker before dispatch
             await self.rctx.add(operation, parameter_set)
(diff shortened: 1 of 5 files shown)


dffml/df/memory.py
```diff
@@ -1408,6 +1349,65 @@ def __init__(
                 else:
                     task.exception()
 
+    async def operations_parameter_set_pairs(
```
pdxjohnny: Did you mean to move this to `MemoryOperationImplementationNetwork` rather than `MemoryOrchestratorContext`? (#452)

aghinsa: @pdxjohnny #452 says is should be in `MemoryOrchestratorContext`,did i miss some disscussion on that?

pdxjohnny: Ah okay i was getting thrown off by how the diff is displayed on GitHub, my bad, at first glance it looked like this got moved under `MemoryOrchestratorContext`.

We should be able to remove the `ictx`, `octx`, `rctx` from the method's arguments and access them via `self.ictx`, etc.

pdxjohnny:
After seeing the LGTM fixed alert I'm reminded that we should remove the method definition from the abstract base class as well: [`BaseOperationImplementationNetworkContext.operations_parameter_set_pairs`](https://lgtm.com/projects/g/intel/dffml/snapshot/8b8c140ec08164c780efeabeaf713bcca7d21efc/files/dffml/df/base.py?sort=name&dir=ASC&mode=heatmap#x8b8512d8a66c199:1)

PR #492, opened 2020-03-26
WIP:Dataflow:Auto start operations without inputs

### LINKED ISSUE #392: df: memory: Dispatch operations without inputs as soon as DataFlow begins
We may have operations which need to be kicked off as soon as a dataflow starts, a good example would be something which is waiting for a message in a gitter channel, and the output it the message. This operation likely has no need for an input, it just returns the first message it sees. We'd want it to start waiting as soon as the dataflow started.

aghinsa:
fix #392


dffml/df/memory.py
```diff
@@ -1026,6 +1026,20 @@ def __init__(
...
         return task
 
+    async def dispatch_auto_starts(self, octx: BaseOrchestratorContext, ctx):
+        """
+        """
+        empty_parameter_set = MemoryParameterSet(
+            MemoryParameterSetConfig(ctx=ctx, parameters=[])
+        )
+
+        for operation in octx.config.dataflow.auto_starts.values():
```
pdxjohnny: I don't think we need the `auto_start` parameter. If an operation doesn't have inputs, then the only way it would get run is if we auto start it. Therefore there's really no need to have a separate property to signify that it needs to be auto started, the lack of inputs will signal that to us.

Also, I don't think this is worth creating a separate data structure for. We only care about this when we first start running a dataflow. So we should probably just skip any operations with inputs when we call this function and dispatch all the ones without.

```suggestion
        for operation in octx.config.dataflow.operations.values():
            if operation.inputs:
                continue
```

aghinsa:
> Great job! :D I left a comment on a way to simplify this. Let's also make sure to add a test.

oh my bad,i thought i forgot to commit the test!

PR #496, opened 2020-03-28
Source: Doctest experiment on Abstract Base Classes.

mHash1m:
Keeping the PR short to see if I'm on the right track with the ABC.

Signed-off-by: Hashim.git <hashimchaudry23@gmail.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `dffml.dataflow.run` operation.
 - Support for operations without inputs.
 - Partial doctestable examples to `features.py`
+- Doctestable examples for `BaseSource`
 ### Fixed
 - New model tutorial mentions file paths that should be edited.
 - DataFlow is no longer a dataclass to prevent it from being exported
--- a/dffml/source/source.py
+++ b/dffml/source/source.py
@@ -29,12 +29,39 @@ class BaseSourceContext(BaseDataFlowFacilitatorObjectContext):
     async def update(self, record: Record):
         """
         Updates a record for a source
+
+        Examples
+        --------
+        >>> async def main():
+        ...     async with MemorySource(records=[]) as source:
+        ...         # Open, update, and close
+        ...         async with source() as ctx:
+        ...             example = Record("one", data=dict(features=dict(feed="face")))
+        ...             # ... Update one into our records ...
+        ...             await ctx.update(example)
+        ...             # Let's check out our records after calling `record` and `update`.
+        ...             async for record in ctx.records():
+        ...                 print(record.export())
+        >>>
+        >>> asyncio.run(main())
+        {'key': 'one', 'features': {'feed': 'face'}, 'extra': {}}
         """
 
     @abc.abstractmethod
     async def records(self) -> AsyncIterator[Record]:
         """
         Returns a list of records retrieved from self.src
+
+        Examples
+        --------
+        >>> async def main():
+        ...     async with MemorySource(records=[Record("example", data=dict(features=dict(dead="beef")))]) as source:
+        ...         async with source() as ctx:
+        ...             async for record in ctx.records():
+        ...                 print(record.export())
+        >>>
+        >>> asyncio.run(main())
+        {'key': 'example', 'features': {'dead': 'beef'}, 'extra': {}}
         """
         # mypy ignores AsyncIterator[Record], therefore this is needed
         yield Record("")  # pragma: no cover
@@ -42,7 +69,24 @@ class BaseSourceContext(BaseDataFlowFacilitatorObjectContext):
     @abc.abstractmethod
     async def record(self, key: str):
         """
-        Get a record from the source or add it if it doesn't exist
+        Get a record from the source or add it if it doesn't exist.
+
+        Examples
+        --------
+        >>> async def main():
+        ...     async with MemorySource(records=[Record("example", data=dict(features=dict(dead="beef")))]) as source:
+        ...         # Open, update, and close
+        ...         async with source() as ctx:
+        ...             example = await ctx.record("example")
+        ...             # Let's also try calling `record` for a record that doesnt exist.
+        ...             one = await ctx.record("one")
+        ...             await ctx.update(one)
+        ...             async for record in ctx.records():
+        ...                 print(record.export())
+        >>>
+        >>> asyncio.run(main())
+        {'key': 'example', 'features': {'dead': 'beef'}, 'extra': {}}
+        {'key': 'one', 'extra': {}}
         """
 
 
--- a/docs/doctest_header.py
+++ b/docs/doctest_header.py
@@ -17,12 +17,13 @@ atexit.register(functools.partial(shutil.rmtree, DOCTEST_TEMPDIR))
 os.chdir(DOCTEST_TEMPDIR)
 
 from dffml_model_scikit import *
-from dffml.record import RecordPrediction, RecordData, Record
 from dffml import *
 from dffml.base import *
+from dffml.record import *
 from dffml.df.base import *
 from dffml.df.types import *
 from dffml.df.memory import *
 from dffml.util.net import *
 from dffml.operation.output import *
 from dffml.operation.dataflow import *
+from dffml.source.memory import *


dffml/source/source.py
```diff
@@ -29,12 +29,45 @@ def __init__(self, parent: "BaseSource") -> None:
...
         """
         Returns a list of records retrieved from self.src
+
+        Examples
+        --------
+        >>> async def main():
+        ...     async with ExampleSource(records=[Record("example")]) as testSource:
+        ...         async with testSource() as sourceContext:
+        ...             ITER = sourceContext.records()
+        ...             async for ITEM in ITER:
```
pdxjohnny: Looking good, let's stay away from uppercase local variables, and change up the names here a bit.

```suggestion
        ...             async for record in ctx.records():
```

dffml/source/source.py
```diff
@@ -29,12 +29,45 @@ def __init__(self, parent: "BaseSource") -> None:
...
     @abc.abstractmethod
     async def records(self) -> AsyncIterator[Record]:
         """
         Returns a list of records retrieved from self.src
+
+        Examples
+        --------
+        >>> async def main():
+        ...     async with ExampleSource(records=[Record("example")]) as testSource:
+        ...         async with testSource() as sourceContext:
```
pdxjohnny:
```suggestion
        ...         async with source() as ctx:
```

dffml/source/source.py
pdxjohnny: Also, let's not make the call to `ctx.records()` a separate line, that is likely to confuse people

docs/api/source/base.rst
```diff
@@ -1,5 +1,6 @@
 Base Source
 ===========
 
+.. literalinclude:: /../docs/references/example_source.py
```
pdxjohnny:
```suggestion
.. literalinclude:: /../docs/references/example_source.py

```

docs/conf.py
```diff
@@ -90,3 +90,10 @@
 doctest_global_setup = (
     pathlib.Path(__file__).parent / "doctest_header.py"
 ).read_text()
+
+doctest_global_setup += (
+    "\n\n"
+    + (
+        pathlib.Path(__file__).parent / "references/example_source.py"
```
pdxjohnny:
```suggestion
        pathlib.Path(__file__).parent / "references" / "example_source.py"
```

We want to make sure we use the [`pathlib.Path` way of joining paths](https://docs.python.org/3/library/pathlib.html#basic-use) (`/` between strings)

docs/references/example_source.py
```diff
@@ -0,0 +1,30 @@
...
+    async def record(self, key: str) -> Record:
+        return self.parent.mem.get(key, Record(key))
+
+
+@config
+class ExampleSourceConfig:
+    records: List[Record]
+
+
+class ExampleSource(BaseSource):
```
pdxjohnny: In this case, you've duplicated `MemorySource`, therefore it's probably be we just use `MemorySource` throughout the docs. We talked about making `Example<PluginType>` classes, however with this one we probably don't need to do that since there's already `MemorySource`.

I realized that we can do the same thing with the model examples. When you go to do the model, let's make sure https://github.com/intel/dffml/issues/500 gets done first, and then you could use that within the docstrings. This way you don't have to make an `Example` class for it.

docs/conf.py
pdxjohnny: We won't need this addition anyway if we just use `MemorySource` (which we should)

dffml/source/source.py
```diff
@@ -43,6 +76,35 @@ def __init__(self, parent: "BaseSource") -> None:
...
+        ...             ITER = sourceContext.records()
+        ...             async for ITEM in ITER:
+        ...                 print(ITEM)
+        ...             one = await sourceContext.record("one")
+        ...             # ... update one ...
+        ...             await sourceContext.update(one)
+        ...             # Let's check out our records after calling `record` and `update`.
+        ...             ITER = sourceContext.records()
+        ...             async for ITEM in ITER:
+        ...                 print(ITEM)
```
pdxjohnny:
```suggestion
        ...                 print(record.export())
```

As you can see record printing has become odd. We should probably avoid it and print the exported version.

dffml/source/source.py
```diff
@@ -43,6 +76,35 @@ def __init__(self, parent: "BaseSource") -> None:
...
+        >>> async def main():
+        ...     async with ExampleSource(records=[Record("example")]) as testSource:
+        ...         # Open, update, and close
+        ...         async with testSource() as sourceContext:
+        ...             # Let's check out our records before calling `record` and `update`.
+        ...             ITER = sourceContext.records()
+        ...             async for ITEM in ITER:
+        ...                 print(ITEM)
+        ...             one = await sourceContext.record("one")
+        ...             # ... update one ...
```
pdxjohnny: We should probably call `record.evaluated` here.

dffml/source/source.py
```diff
@@ -29,12 +29,45 @@ def __init__(self, parent: "BaseSource") -> None:
...
         Updates a record for a source
+
+        Examples
+        --------
+        >>> async def main():
+        ...     async with ExampleSource(records=[]) as testSource:
+        ...         # Open, update, and close
+        ...         async with testSource() as sourceContext:
+        ...             example = Record("one")
+        ...             # ... Update one into our records ...
```
pdxjohnny: We should probably call `record.evaluated` here.

dffml/source/source.py
```diff
@@ -29,12 +29,45 @@ def __init__(self, parent: "BaseSource") -> None:
...
         Returns a list of records retrieved from self.src
+
+        Examples
+        --------
+        >>> async def main():
+        ...     async with ExampleSource(records=[Record("example")]) as testSource:
+        ...         async with testSource() as sourceContext:
+        ...             ITER = sourceContext.records()
+        ...             async for ITEM in ITER:
+        ...                 print(ITEM)
```
pdxjohnny:
```suggestion
        ...                 print(record.export())
```

As you can see record printing has become odd. We should probably avoid it and print the exported version.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add python code for tensorflow DNNEstimator
 - Ability to run a subflow as if it were an operation using the
   `dffml.dataflow.run` operation.
+- Doctestable examples for an Abstract Base Class in `Base Source`.
```
pdxjohnny:
```suggestion
- Doctestable examples for `BaseSource`.
```

pdxjohnny: We probably don't need to call out that it's an Abstract Base Class, also there's no space

dffml/source/source.py
mHash1m: Yes, it had been itching at me. Thank you.

dffml/source/source.py
mHash1m: Yeah, thanks. I've added data to `example` and also added another record using ctx.record since it can also add new records if they dont exist previously. I'm leaving it without data to show that it simply adds a record without any data.

dffml/source/source.py
mHash1m: Evaluated will cause records.export() to leave a last updated timestamp here, do we want that? I'm not sure how to include timestamps in sphinx doctests but I can surely figure something out if that's the case, if not I can just add features through the initializer.

dffml/source/source.py
mHash1m: Also, do we want "extra" to not show as empty?

docs/references/example_source.py
mHash1m: Well yes, it was bugging me too. 
Sure thing! Thank you.

mHash1m:
Thank you for going through all that! I'm get a lot to learn from your reviews!

dffml/source/source.py
pdxjohnny: Ah yes, the timestamps... Maybe we should just call `record.features()` then.

dffml/source/source.py
pdxjohnny: `extra` is meant as an area where sources can store other information they might need on a record to `update` it properly. It's not relevant here, no need to show any data in it.

pdxjohnny:
Thanks @mHash1m! Great stuff! I had forgotten to hit merge on this, Merged! :D

PR #506, opened 2020-03-30
Update README.md

### LINKED ISSUE #504: docs: Update README for shouldi
https://github.com/intel/dffml/blob/master/examples/shouldi/README.md the docs need to be updated as `shouldi` has evolved since. There are many  new code analyzers wrapped by `shouldi ` which are missing in the README for example, npm_audit, go lang's linter etc. The TODO's mentioned are actually done .

TakshKamlesh:
FIxes #504 
Kindly inform what more could be updated / removed.

### DIFF
--- a/examples/shouldi/README.md
+++ b/examples/shouldi/README.md
@@ -28,12 +28,8 @@ any issues in any of the python packages you were thinking of installing.
 Right now `shouldi` runs the following static analysis tools and complains if:
 
 - [safety](https://pyup.io/safety/)
-  - Any issues are found
   
-- TODO: [pylint](https://pypi.org/project/pylint/)
-  - TDB (something about the number of errors)
-  
-Also , shouldi as been updated with following code wrappers:
+shouldi as been updated with following code wrappers:
 
 - [bandit](https://pypi.org/project/bandit/)
 - [golangci_lint](https://github.com/golangci/golangci-lint/blob/master/README.md)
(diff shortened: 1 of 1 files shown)


examples/shouldi/README.md
```diff
@@ -29,9 +29,15 @@ Right now `shouldi` runs the following static analysis tools and complains if:
 
 - [safety](https://pyup.io/safety/)
   - Any issues are found
-- TODO: [bandit](https://pypi.org/project/bandit/)
+  
 - TODO: [pylint](https://pypi.org/project/pylint/)
```
pdxjohnny, line 33: Let's remove the bullet points about `pylint`

examples/shouldi/README.md
```diff
@@ -29,9 +29,15 @@ Right now `shouldi` runs the following static analysis tools and complains if:
...
   - TDB (something about the number of errors)
+  
+Also , shouldi as been updated with following code wrappers:
```
pdxjohnny, line 36: - I don't think we need to say `Also, ...` here. We can just list all of them.
- Let's also remove the `Any issues are found` line below the `safety` bullet point

TakshKamlesh: OK I am working on it

TakshKamlesh: Done

TakshKamlesh:
What's the issue with changelog file

PR #516, opened 2020-04-01
operation: Adding new operations

0dust:
Added new operations `AcceptUserInput` (takes input from the user using python input()) and `print_output` (prints using python print()).


tests/operation/test_io.py
```diff
@@ -0,0 +1,69 @@
...
+from dffml.operation.output import GetSingle
+from dffml.df.memory import MemoryOrchestrator
+from dffml.util.asynctestcase import AsyncTestCase
+from dffml.df.types import DataFlow, Input, Definition
+from dffml.operation.io import AcceptUserInput, printOutput
+
+
+class TestInputOutput(AsyncTestCase):
+    async def setUp(self):
+        self.stdout = io.StringIO()
```
pdxjohnny, line 16: I can't remember, do we need to call `super().setUp()` here? If there's a `setUp()` in AsyncTestCase then we should. If not then don't worry about it.

pdxjohnny: Other than that this file looks great..

0dust: Yes there is `setUp()` in AsyncTestCase.

dffml/operation/io.py
```diff
@@ -0,0 +1,66 @@
...
+        self.__pool.__exit__(exc_type, exc_value, traceback)
+        self.__pool = None
+        self.pool = None
+        self.loop = None
+
+
+@op(
+    inputs={"DataToPrint": DataToPrint}, outputs={}, conditions=[],
+)
+async def printOutput(DataToPrint: str):
```
pdxjohnny, line 65:
```suggestion
async def print_output(data: str):
```

I just realized that this breaks with convention. Functions and arguments should be underscored. Only class names should be CamelCase (This is just a standard Python thing, not specific to DFFML).

I also would recommend naming the argument `data` rather than `DataToPrint`, since then you'll just end up with a variable and class named very similarly.

pdxjohnny:
Alright last thing @0dust, sorry today was hectic for me, I think we should add some examples for these operations, similar to the example usage for the `dffml/operation/dataflow.py:run_dataflow` operation. Also, we need to make sure they get listed in the top level `setup.py` file. Thanks!

dffml/operation/io.py
```diff
@@ -36,6 +35,35 @@ def receive_input():
...
+    Accept input from stdin using python input()
+    Parameters
+    ++++++++++
+    inputs : dict
+        A dictionary with a key and empty list as value.
+
+    Returns
+    +++++++
+    dict
+        A dictionary containing user input.
```
pdxjohnny, line 48:
```suggestion
        A dictionary containing user input.

```

dffml/operation/io.py
```diff
@@ -36,6 +35,35 @@ def receive_input():
 
 
 class AcceptUserInput(OperationImplementation):
+    """
+    Accept input from stdin using python input()
```
pdxjohnny, line 39:
```suggestion
    Accept input from stdin using python input()

```

dffml/operation/io.py
```diff
@@ -36,6 +35,35 @@ def receive_input():
...
+    ++++++++++
+    inputs : dict
+        A dictionary with a key and empty list as value.
+
+    Returns
+    +++++++
+    dict
+        A dictionary containing user input.
+    Examples
+    ++++++++
```
pdxjohnny, line 50:
```suggestion
    ++++++++

```

pdxjohnny:
Awesome work with this!!

PR #518, opened 2020-04-02
WIP:dataflow:Async generator operations

### LINKED ISSUE #394: df: memory: async iterator operations
Functions which are async iterators (`async def` with `yield`) should be supported by `@op`, as well as `dffml/df/memory.py` needs to be able to run one of these kinds of operations. Everytime the operation `yield`s, whatever it `yield`'d should be added to the input network context.

I think we'll likely want to combine `run` and `run_dispatch` so that we only have the `run` function, it's looking redundant (at first glance) *whoever submits a PR please figure this out and consolidate if appropriate*

https://github.com/intel/dffml/blob/7dfec5050c0e2f205fdd58524a420b861dc7fd9e/dffml/df/memory.py#L861-L976

aghinsa:
fix #394

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Python prompts `>>>` can now be enabled or disabled for easy copying of code into interactive sessions.
 - `GetMulti` operation which gets all Inputs of a given definition
+- Support for async generator operations
+
 ### Changed
 - Renamed `"arg"` to `"plugin"`.
 
@@ -31,7 +33,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for operations without inputs.
 - Partial doctestable examples to `features.py`
 - Doctestable examples for `BaseSource`
-- Support for async generator operations
 - Instructions for setting up debugging environment in VSCode
 ### Fixed
 - New model tutorial mentions file paths that should be edited.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Python prompts `>>>` can now be enabled or disabled for easy copying of code into interactive sessions.
 - `GetMulti` operation which gets all Inputs of a given definition
 - Support for async generator operations
-
 ### Changed
 - Renamed `"arg"` to `"plugin"`.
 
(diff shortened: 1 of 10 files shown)


dffml/df/memory.py
```diff
@@ -1431,12 +1437,11 @@ def __init__(
...
         # Create initial events to wait on
         # TODO(dfass) Make ictx.added(ctx) specific to dataflow
-        input_set_enters_network = asyncio.create_task(self.ictx.added(ctx))
-        tasks.add(input_set_enters_network)
         # schedule running of operations with no inputs
         async for task in self.nctx.dispatch_auto_starts(self, ctx):
             tasks.add(task)
-
+        input_set_enters_network = asyncio.create_task(self.ictx.added(ctx))
+        tasks.add(input_set_enters_network)
```
pdxjohnny, line 1444: Was there a reason for moving this?

tests/df/test_async_iter.py
```diff
@@ -0,0 +1,81 @@
...
+
+@op(inputs={}, outputs={"count": Count})
+async def counter_auto_start():
+    for i in range(5):
+        yield {"count": i}
+
+
+@op(inputs={"number_in": Count}, outputs={"number_out": Number})
+def echo_num(number_in: int):
+    print(f"{number_in}")
```
pdxjohnny, line 28:
```suggestion
```

dffml/df/memory.py
pdxjohnny: Ah, I see your Gitter message. That's not good if we don't know why :/

aghinsa: Yeah,been trying to figure it out.I was trying some stuff and this swapping was a total fluke.

pdxjohnny:
@aghinsa I added GetMulti (https://github.com/intel/dffml/pull/525) if you could use it to grab all the counts and confirm that they're all there then we can merge this. Thanks!

dffml/df/memory.py
aghinsa: The operations without inputs are dispatched outside the loop of other operations,and since earlier they were added after `input_set_enters_network` ,the network was not getting notified of the outputs from the async gen.
^ Reason for the swap

CHANGELOG.md
```diff
@@ -30,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for operations without inputs.
 - Partial doctestable examples to `features.py`
 - Doctestable examples for `BaseSource`
+- Support for async generator operations
```
pdxjohnny, line 34: We need to move this up to the `Unreleased` section

PR #528, opened 2020-04-04
operation: io: Add io operations demo

0dust:


examples/io/io_usage.py
```diff
@@ -0,0 +1,63 @@
...
+        DefFeature("Expertise", int, 1),
+        DefFeature("Trust", float, 1),
+    ),
+    predict=DefFeature("Salary", int, 1),
+)
+
+# TODO Training should be in different file, only prediction
+# should be here?
+# TODO decide on the stdin input format, if its like {"Years": 0, "Expertise": 1, "Trust": 0.1, "Salary": 10}
+# then literal_eval or something.. ask John
```
pdxjohnny, line 25: - Make another operation, `dffml/operation/preprocess.py` for the `literal_eval` operation.
- For the flow, connect `modelprediction.features` to `dffml.mapping.create`
  - Connect `dffml.mapping.create`  value to `literal_eval`
  - Add to dataflow.seed, an `Input` with the `origin="seed.feature"`
```python
        "create_feature_map": InputFlow(
            inputs={"key": ["seed.feature"], ... literal_eval connection for value }
        ),
```

- Change the demo to use the slr model in `dffml/model/slr.py`
  - That way it only needs one piece of input feature data.

docs/usage/io.rst
```diff
@@ -0,0 +1,40 @@
+Prediction Using IO Operations
+==============================
+
+This example will show you how to train a model using DFFML python API and use the
+model for prediction by taking input from stdio.
+
+The model we'll be using is ``slr`` which is a part of ``dffml`` models.
```
pdxjohnny, line 7: It's probably sufficient to say that `slr`  comes from `dffml` and leave off the `pip install` command. Since it's in the base package, we can assume they've already installed `dffml`.

```suggestion
We'll be using the Simple Linear Regression model
(:ref:`plugin_model_dffml_slr`) which is in the ``dffml`` package.
```

docs/usage/io.rst
```diff
@@ -0,0 +1,40 @@
...
+
+
+.. literalinclude:: /../examples/io/io_usage.py
+
+On running the above code `AcceptUserInput` operation
+waits for input from stdio.
+
+.. code-block:: console
+
+    Enter the value:
```
pdxjohnny, line 26: Change to keep consistent with `end=""` change to `print` above.

```suggestion
    Enter the value: 21
```

dffml/operation/io.py
```diff
@@ -25,13 +25,14 @@
 class AcceptUserInputContext(OperationImplementationContext):
     @staticmethod
     def receive_input():
+        print("Enter the value:")
```
pdxjohnny, line 28:
```suggestion
        print("Enter the value: ", end="")
```

This way the user can type in input on the same line

docs/usage/io.rst
```diff
@@ -0,0 +1,40 @@
...
+    21
```
pdxjohnny, line 27:
```suggestion
```

docs/usage/io.rst
```diff
@@ -0,0 +1,40 @@
...
+
+The value (here "21") is then converted to `int` by
```
pdxjohnny, line 29: We should explicitly say that it's a `str`

```suggestion
The value (which is the `str` "21") is then converted to `int` by
```

docs/usage/io.rst
```diff
@@ -0,0 +1,40 @@
...
+`literal_eval` operation.
+A mapping is created by `create_mapping` operation ({"Years": 21})
```
pdxjohnny, line 31:
```suggestion
A mapping (`dict`) is created by `create_mapping` operation (`{"Years": 21}`)
```

docs/usage/io.rst
```diff
@@ -0,0 +1,40 @@
...
+which is then passed to `model_predict` operation for making prediction.
```
pdxjohnny, line 32: We should say something about the reason why we create a mapping is because `model_predict` takes a mapping for feature name to feature value.

examples/io/io_usage.py
```diff
@@ -0,0 +1,78 @@
...
+        value="Years",
+        definition=create_mapping.op.inputs["key"],
+        origin="seed.Years",
+    )
+)
+
+
+async def main():
+    async for ctx, results in MemoryOrchestrator.run(dataflow, {"inputs": []}):
+        print("Finished")
```
pdxjohnny, line 75:
```suggestion
        pass
```

We probably don't need it to say finished, the program will exit when it finishes anyway.

docs/usage/io.rst
```diff
@@ -0,0 +1,36 @@
+Prediction Using IO Operations
+==============================
+
+This example will show you how to train a model using DFFML python API and use the
+model for prediction by taking input from stdio.
+
+DFFML offers several :ref:`plugin_models`. For this example
+we will be using the Simple Linear Regression model
+( ``slr`` ) which is in the ``dffml`` package.
+
```
0dust, line 10: @pdxjohnny Putting link to `Models` page. 
:ref:`plugin_model_dffml_slr` doesn't seem to work. I tried putting an anchor for `slr` to reference it but am unable to do it because `dffml_model.rst` gets overwritten. What am I missing?

pdxjohnny: Have you merged master? I just added the references recently

docs/usage/io.rst
```diff
@@ -0,0 +1,36 @@
...
+
+.. literalinclude:: /../examples/io/io_usage.py
+
+On running the above code `AcceptUserInput` operation
+waits for input from stdio.
+
+.. code-block:: console
+
+    Enter the value: 21
```
pdxjohnny, line 22: Below this code block would be a good place for a dataflow diagram.

https://intel.github.io/dffml/cli.html#diagram

```console
$  dffml service dev export examples.io.io_usage:dataflow | dffml dataflow diagram -config json /dev/stdin
```

You'll paste that into the mermaidjs live editor and then you'll want to save the `.svg`.

But first I'll need to merge https://github.com/intel/dffml/pull/552 because export is currently broken.

Also, you'll need to apply this patch

```diff
diff --git a/examples/io/io_usage.py b/examples/io/io_usage.py
index c3077dc3..d8182a50 100644
--- a/examples/io/io_usage.py
+++ b/examples/io/io_usage.py
@@ -1,7 +1,6 @@
 import asyncio

-from dffml.noasync import train
-from dffml import Features, DefFeature
+from dffml import train, Features, DefFeature
 from dffml.operation.output import GetSingle
 from dffml.df.memory import MemoryOrchestrator
 from dffml.operation.mapping import create_mapping
@@ -16,15 +15,6 @@ slr_model = SLRModel(
     predict=DefFeature("Salary", int, 1),
 )

-# Train the model
-train(
-    slr_model,
-    {"Years": 0, "Salary": 10},
-    {"Years": 1, "Salary": 20},
-    {"Years": 2, "Salary": 30},
-    {"Years": 3, "Salary": 40},
-)
-
 # This Dataflow takes input from stdio using `AcceptUserInput`
 # operation. The string input which corresponds to feature `Years`
 # is converted to `int`/`float` by
@@ -71,8 +61,18 @@ dataflow.seed.append(


 async def main():
+    # Train the model
+    await train(
+        slr_model,
+        {"Years": 0, "Salary": 10},
+        {"Years": 1, "Salary": 20},
+        {"Years": 2, "Salary": 30},
+        {"Years": 3, "Salary": 40},
+    )
+
     async for ctx, results in MemoryOrchestrator.run(dataflow, {"inputs": []}):
         pass


-asyncio.run(main())
+if __name__ == "__main__":
+    asyncio.run(main())
```

Or else when you try to export it will try to run `train`. We should include instructions on generating the diagram in the tutorial as well.

docs/usage/io.rst
0dust: Ohh! Thanks! Done.

pdxjohnny:
Great work @0dust!!! :D

PR #529, opened 2020-04-05
docs: hiding of python prompts

### LINKED ISSUE #438: docs: Enable hiding of Python prompts
This will be very helpful for copy pasting examples.

References:
- https://github.com/readthedocs/sphinx_rtd_theme/issues/167

sk-ip:
The python prompt `>>>` can be enabled or disabled. this way users can easily copy paste code in interactive sessions.
closes #438

### DIFF
--- a/scripts/docs.sh
+++ b/scripts/docs.sh
@@ -11,5 +11,8 @@ python3.7 scripts/docs.py
 python3.7 -c 'import os, pkg_resources; [e.load() for e in pkg_resources.iter_entry_points("console_scripts") if e.name.startswith("sphinx-build")][0]()' -b html docs pages \
   || (echo "[ERROR] Failed run sphinx, is it installed (pip install -U .[dev])?" 1>&2 ; exit 1)
 cp -r docs/images pages/
-wget -P pages/_static "https://raw.githubusercontent.com/python/python-docs-theme/master/python_docs_theme/static/copybutton.js"
+curl -sSL -o pages/_static/copybutton.js "https://raw.githubusercontent.com/python/python-docs-theme/master/python_docs_theme/static/copybutton.js"
+sha384sum -c - <<EOF
+bf80c778867bd87d14588ff72ae10632b331427379a299ab2ac4d7ddefa9b648313720b796ab441359e0e47daf738109 pages/_static/copybutton.js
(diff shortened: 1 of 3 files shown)


sk-ip:
@pdxjohnny , the only left is to add `copybutton.js` file to the `static` directory. but the `static` directory cannot be found. i found one in the `pages` directory after running `./scripts/docs.sh` and it worked fine. but i don't think its the best place to add that file, since it is also getting ignored. So where should i add the `copybutton.js` file.

pdxjohnny:
We want `./scripts/docs.sh` to download it from https://github.com/python/python-docs-theme/raw/master/python_docs_theme/static/copybutton.js and then [`patch`](https://manpage.me/?q=patch) with any needed adjustments for the read the docs theme.

sk-ip:
@pdxjohnny , i added a link to download the `copybutton.js` file to the path `pages/_static` and then ran `./scripts.docs.sh` and it worked : )
here's how it looks

ezgif com-optimize

sk-ip:
@pdxjohnny , this is ready for review. i don;t know why some tests are failing.

scripts/docs.sh
```diff
@@ -11,4 +11,5 @@ python3.7 scripts/docs.py
 python3.7 -c 'import os, pkg_resources; [e.load() for e in pkg_resources.iter_entry_points("console_scripts") if e.name.startswith("sphinx-build")][0]()' -b html docs pages \
   || (echo "[ERROR] Failed run sphinx, is it installed (pip install -U .[dev])?" 1>&2 ; exit 1)
 cp -r docs/images pages/
+wget -P pages/_static "https://raw.githubusercontent.com/python/python-docs-theme/master/python_docs_theme/static/copybutton.js"
```
pdxjohnny, line 14: Nice!!! Can we pin this to a hash and using `sha384sum` to verify the contents are what we expect? The [MNIST](https://intel.github.io/dffml/usage/mnist.html) demo has an example of how to verify `sha384sum`s. [dffml.util.net.cached_download](https://intel.github.io/dffml/api/util/net.html#dffml.util.net.cached_download) has an example of how to calculate the hash you want to use for comparison.

PR #541, opened 2020-04-08
docs: model: scratch: Python code examples

sk-ip:
Python code examples for using dffml_model_scratch.
references #434

PR #542, opened 2020-04-08
WIP: docs: model: tensorflow hub: Include examples of how to use from Python API

### LINKED ISSUE #435: docs: model: tensorflow hub: Include examples of how to use from Python API
The model plugins page has command line examples but needs examples from Python as well.

Copy the format of the `examples/quickstart_filenames.py` file and add a `.. code-block:: python` section for each model, keeping the Python example consistent with the command line example for each model. Let's add the python example after the CLI example for each.

For the tensorflow hub model, the docstring of the `Model` needs to be updated.

**Edit:**

Here's a good example from when tensorflow was updated: https://github.com/intel/dffml/commit/e7aaeb1e1e2f1b2ce74d1fe6ecc6422bf6df178c

TakshKamlesh:
Fixes #435

### DIFF
--- a/model/tensorflow_hub/dffml_model_tensorflow_hub/__init__.py
+++ b/model/tensorflow_hub/dffml_model_tensorflow_hub/__init__.py
@@ -1,0 +1,51 @@
+"""
+
+from dffml import CSVSource, Features, DefFeature
+from dffml.noasync import train, accuracy, predict
+from dffml_model_tensorflow_hub.text_classifier import TextClassificationModel
+
+
+model = TextClassificationModel(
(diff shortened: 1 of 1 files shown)


pdxjohnny:
Nice! We also need to make sure we can test the example.

https://github.com/intel/dffml/commit/e7aaeb1e1e2f1b2ce74d1fe6ecc6422bf6df178c

is a good example of how to make this testable, you can ignore everything in `docs/plugins/dffml_model.rst`

pdxjohnny:
https://github.com/intel/dffml/pull/541 is also a good example

PR #555, opened 2020-04-12
source: ini: Source for parsing .ini files

### LINKED ISSUE #551: source: ini: Create source for parsing .ini files
The goal of this issue is to show how to create a source derived from `FileSource` and `MemorySource` which saves and loads data from `.ini` files.

To do this we will create a new source, `INISource` (the call to `entrypoint` will be `@entrypoint("ini")`).

The source will be similar in structure to the `IDX1` and `IDX3` sources. Only we'll be implementing the `dump_fd` method in this one.

The [`configparser`](https://docs.python.org/3/library/configparser.html) library will be helpful here, since it parses `.ini` files.

Since we're working with file objects in `load_fd` and `dump_fd`, rather than filenames, we'll want to use the [`read_file`](https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.read_file) method within `load_fd` .

In `dump_fd`, we'll likely want to use [`read_dict`](https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.read_dict) to create a `Configparser`. Both `load_fd` and `dump_fd` should have local variables which are an instance of `Configparser`. We don't want to make it class local because if features in an individual record are deleted then they would still show up in the output.

You'll want to make sure to call `record.features()` and using that as the data to save for the record. Also assume when loading that you're loading the feature data. Each `[section]` should be used as a record. With the name of the section being the `key` for that record.

Make sure to write lots of comments, since this will be used as the example for another tutorial on Sources (see issue https://github.com/intel/dffml/issues/282).

Here is the `idx1` source as an example.

https://github.com/intel/dffml/blob/155c114da35b80d5a17c96eece5c45dc8a6322c5/dffml/source/idx1.py#L6-L56

sk-ip:
Add source for parsing .ini file formats.
references #551 closes #551


pdxjohnny:
Looking great! Next step, write `dump_fd`, then copy the format of the testcase for `JSON` https://github.com/intel/dffml/blob/master/tests/source/test_json.py

dffml/source/ini.py
```diff
@@ -0,0 +1,65 @@
...
+                    data={"features": {str(name): str(value)}},
+                )
+
+        self.logger.debug("%r loaded %d sections", self, len(self.mem))
+
+    async def dump_fd(self, fd):
+        # create an instance of configparser
+        parser = ConfigParser()
+        # read the fileobj(fd) as dict where key is section name and values are dict of option and value pair
+        parser.read_dict(fd)
```
pdxjohnny, line 51: You don't want this line

```suggestion
```

dffml/source/ini.py
```diff
@@ -0,0 +1,65 @@
...
+
+        # go over each section in mem
+        for section in self.mem.keys():
```
pdxjohnny, line 54: `.keys()` will just give you the section headers (aka `record.key`), we want to use `.items()` to get the record itself.

```suggestion
        for section, record in self.mem.items():
```

dffml/source/ini.py
```diff
@@ -0,0 +1,65 @@
...
+            # get each section data as a dict
+            section_data = section.features()
```
pdxjohnny, line 56: The feature data will be in the record itself

```suggestion
            section_data = record.features()
```

dffml/source/ini.py
```diff
@@ -0,0 +1,63 @@
...
+from .file import FileSource
+from ..base import config, field
+from .memory import MemorySource
+from ..util.entrypoint import entrypoint
+
+
+@config
+class INISourceConfig:
+    filename: str
+    feature: str = field("Name of the feature the data will be referenced as")
```
pdxjohnny, line 13: I don't think to meant to keep this line (the GitHub actions logs are yelling about it)

```suggestion
```

pdxjohnny:
And don't forget to add to the changelog ;)

tests/source/test_ini.py
```diff
@@ -0,0 +1,12 @@
...
+from dffml.util.asynctestcase import AsyncTestCase
+
+
+class TestINISource(FileSourceTest, AsyncTestCase):
+    async def setUpSource(self):
+        return INISource(
+            INISourceConfig(
+                filename=self.testfile, allowempty=True, readwrite=True
+            )
+        )
```
pdxjohnny, line 12:
```suggestion
        )

    @unittest.skip("tags not implemented")
    async def test_tag(self):
        """
        tags not implemented
        """
```

tests/source/test_ini.py
```diff
@@ -0,0 +1,12 @@
+from dffml.source.ini import INISource, INISourceConfig
+from dffml.util.testing.source import FileSourceTest
+from dffml.util.asynctestcase import AsyncTestCase
+
+
+class TestINISource(FileSourceTest, AsyncTestCase):
+    async def setUpSource(self):
```
pdxjohnny, line 7: Don't involve `FileSourceTest`, just write a couple little tests with comments that test the basic functionality

tests/source/test_ini.py
```diff
@@ -0,0 +1,58 @@
...
+                "section2": Record(
+                    "section2", data={"features": {"C": 3, "D": 4,}}
+                ),
+            }
+            source.mem = dump_data
+            with open(self.testfile, "w") as tfile:
+                await source.dump_fd(tfile)
+
+            allrecords = [record async for record in load(source)]
+            print(allrecords)
```
sk-ip, line 58: @pdxjohnny , here while dumping all the data to the file object, before the dump is complete the load(source) code gets called and only some of the records get loaded. this is happening because of the async nature of load() function. So i am not able to test properly what all things are loaded and dumped.

tests/source/test_ini.py
```diff
@@ -0,0 +1,58 @@
...
+                readwrite=True,
+            )
+            # Save some data in the source
+            await save(
+                source,
+                Record("section1", data={"features": {"A": 1, "B": 2,}}),
+                Record("section2", data={"features": {"C": 3, "D": 4,}}),
+            )
+            # Load the data
+            source.load("ini")
```
pdxjohnny, line 27:
```suggestion
```

tests/source/test_ini.py
```diff
@@ -0,0 +1,58 @@
...
+                allowempty=True,
+                readwrite=True,
+            )
+            # Save some data in the source
+            await save(
+                source,
+                Record("section1", data={"features": {"A": 1, "B": 2,}}),
+                Record("section2", data={"features": {"C": 3, "D": 4,}}),
+            )
+            # Load the data
```
pdxjohnny, line 26: This is not necessary

```suggestion
```

tests/source/test_ini.py
pdxjohnny: You probably only need one testcase. First call save, then call load and confirm that the records came back as you expected. Similarly to how you did `assertDictEqual`

sk-ip:
```
FAIL: test_load_fd (tests.source.test_ini.TestINISource)
  File "/home/runner/work/dffml/dffml/dffml/util/asynctestcase.py", line 66, in run_it
  File "/opt/hostedtoolcache/Python/3.7.6/x64/lib/python3.7/asyncio/base_events.py", line 583, in run_until_complete
    return future.result()
  File "/home/runner/work/dffml/dffml/tests/source/test_ini.py", line 29, in test_load_fd
    self.assertDictEqual(records[0].features(), {"A": 1, "B": 2})
AssertionError: {'b': '2'} != {'A': 1, 'B': 2}
- {'b': '2'}
+ {'A': 1, 'B': 2}

```
@pdxjohnny  Here's the test which was failing, somehow all the records are not loaded properly.

sk-ip:
@pdxjohnny , just found that `configparser` parse the data in a case-insensitive manner https://docs.python.org/3/library/configparser.html#mapping-protocol-access , hence i changed the assert conditions to suit it.

tests/source/test_ini.py
```diff
@@ -0,0 +1,30 @@
...
+                source,
+                Record("section1", data={"features": {"A": 1, "B": 2,}}),
+                Record("section2", data={"features": {"C": 3, "D": 4,}}),
+            )
+            # Load all the records
+            records = [record async for record in load(source)]
+
+            self.assertIsInstance(records, list)
+            self.assertEqual(len(records), 2)
+            self.assertDictEqual(records[0].features(), {"a": "1", "b": "2"})
```
pdxjohnny, line 29:
```suggestion
            self.assertDictEqual(records[0].features(), {"a": 1, "b": 2})
```

tests/source/test_ini.py
```diff
@@ -0,0 +1,30 @@
...
+            self.assertDictEqual(records[1].features(), {"c": "3", "d": "4"})
```
pdxjohnny, line 30:
```suggestion
            self.assertDictEqual(records[1].features(), {"c": 3, "d": 4})
```

PR #556, opened 2020-04-12
Removed Unused import

Yash-Varshney:


Yash-Varshney:
@pdxjohnny  what just happened here ? I expected it to work smooth as I have removed few unused imports .

pdxjohnny:
@Yash-Varshney It looks like some of the imports you removed we're false positives, they are actually used. Next steps here would be to read the GitHub actions logs and see which ones need to be added back in. Then, for the ones that you had to add back in but LGTM thinks should be removed. You'll want to figure out how what the correct comment is to add to the line causing the false positive. LGTM will have a way to ignore issues for a line.

lgtm.yaml
```diff
@@ -3,6 +3,8 @@ path_classifiers:
     - scripts/docs.py
   test:
     - tests
+  queries:
+    - exclude: py/import-and-import-from
```
pdxjohnny, line 7: This makes it so all unused imports are ignored. You want to remove this, and add comments on the individual lines themselves which were false positives.

image

Yash-Varshney: @pdxjohnny this actually does not ignore all the unused imports, in LGTM there is an error which is raised when we use something like - 
` import pathlib
from pathlib import Path`
this error will be raised even when we use `Path`
LGTM asks us to do the above thing like this - 
`Remove the from xxx import yyy statement. Add yyy = xxx.yyy if required.`

Should I still make a comment on particular file or is it okay to exclude these type of errors.

pdxjohnny: Ah yes. We still don't want to ignore those, we should be fixing them instead

pdxjohnny:
Thanks Yash!!!

Issue #325, opened 2020-01-15
util: cli: Unify configuration


pdxjohnny:
Right now the code for parsing command line arguments in `dffml/base.py` and `dffml/util/cli/`. We should get rid of the `arg_` notation when defining CLI commands and replace it with the `CONFIG = ExampleConfig` style syntax (#164). This will probably involve moving that config parsing code from `dffml/base.py` into util, somewhere like `dffml/util/config.py` (or something), and then using it in `dffml/util/cli/`.

pdxjohnny:
After we merge this we'll want to get rid of position, because we realized that the information on position is just the same as the index of the field within the config class.

PR #569, opened 2020-04-16
util: cli: Unify cli Configuration

sakshamarora1:
Fixes: #325

Signed-off-by: Saksham Arora sakshamarora1001@gmail.com


dffml/cli/dataflow.py
```diff
@@ -92,9 +96,29 @@ async def run(self):
...
+        default_factory=lambda: Sources(
+            JSONSource(
+                FileSourceConfig(
+                    filename=os.path.join(
+                        os.path.expanduser("~"), ".cache", "dffml.json"
+                    )
+                )
+            )
+        ),
+        metadata= {
```
pdxjohnny, line 113: Since these are going to be very common, let's have field put them in metadata for us to shorten the amount of code we're writing

```python

def field(description: str, *args, required: bool = False, ... (position and labeled here too) ... , metadata: Optional[dict] = None, **kwargs):
    """
    Creates an instance of :py:func:`dataclasses.field`. The first argument,
    ``description`` is the description of the field, and will be set as the
    ``"description"`` key in the metadata ``dict``.
    """
    if not metadata:
        metadata = {}
    #  Add code for positional arguments
    metadata["description"] = description
    metadata["required"] = required
    return dataclasses.field(*args, metadata=metadata, **kwargs)
```

dffml/util/cli/cmd.py
```diff
@@ -87,21 +89,32 @@ def add_subs(self, add_from: "CMD"):
                 except argparse.ArgumentError as error:
                     raise Exception(repr(add_from)) from error
```
pdxjohnny, line 91: TODO, only call `self.add_argument` using args that don't have a `position` in this loop, append ones with a position to an array or dict, after the loop, sort it by position, and then make another loop where you add them in the correct order

dffml/util/cli/cmds.py
```diff
@@ -70,31 +72,51 @@ def __init__(self, *args, **kwargs):
                 self.sources[i] = self.sources[i].withconfig(self.extra_config)
 
 
+
+@config
+class ModelCMDConfig(CMDConfig):
+    model: Model = field(
+        "Model used for ML",
+        default_factory = Model.load,
```
pdxjohnny, line 80: This shouldn't have a default. It should be required

```suggestion
        required=True,
```

dffml/util/cli/cmds.py
```diff
@@ -70,31 +72,51 @@ def __init__(self, *args, **kwargs):
...
         super().__init__(*args, **kwargs)
         if inspect.isclass(self.model):
             self.model = self.model.withconfig(self.extra_config)
 
 
+@config
+class PortCMDConfig(CMDConfig):
+    port: Port = field(
+        "Port",
+        default_factory = Port.load,
```
pdxjohnny, line 100:
```suggestion
```

dffml/util/cli/cmds.py
```diff
@@ -70,31 +72,51 @@ def __init__(self, *args, **kwargs):
...
 
-    arg_port = Arg("port", type=Port.load)
+    CONFIG = PortCMDConfig
+
+
+@config
+class KeysCMDConfig(CMDConfig):
+    keys: Port = field(
+        "Key used for source lookup and evaluation.",
+        default_factory = Port.load,
```
pdxjohnny, line 113:
```suggestion
```

dffml/util/cli/cmds.py
```diff
@@ -70,31 +72,51 @@ def __init__(self, *args, **kwargs):
...
+
 class PortCMD(CMD):
 
-    arg_port = Arg("port", type=Port.load)
+    CONFIG = PortCMDConfig
+
+
+@config
+class KeysCMDConfig(CMDConfig):
+    keys: Port = field(
```
pdxjohnny, line 111:
```suggestion
    keys: str = field(
```

dffml/util/cli/cmd.py
```diff
@@ -87,21 +89,32 @@ def add_subs(self, add_from: "CMD"):
...
+                try:
+                    # Add condition for positional arguments
+                    self.add_argument("-" + field.name, **arg)
+                except argparse.ArgumentError as error:
+                    raise Exception(repr(add_from)) from error
+
+
+@config
+class CMDConfig:
+    log: str = field(
```
pdxjohnny, line 104:
```suggestion
    log: int = field(
```

dffml/util/cli/cmd.py
```diff
@@ -87,21 +89,32 @@ def add_subs(self, add_from: "CMD"):
...
+        "Logging Level",
+        default=logging.INFO,
+        metadata={"action": ParseLoggingAction, "required": False,},
```
pdxjohnny, line 107:
```suggestion
        action=ParseLoggingAction,
        required=False
```

dffml/cli/dataflow.py
sakshamarora1: Okay!

dffml/util/cli/cmds.py
sakshamarora1: Not providing a default argument here throws a `TypeError: non-default argument 'model' follows default argument` because of `ModelCMDConfig` inheriting from `CMDConfig` which has log with a default value.

dffml/util/cli/cmds.py
sakshamarora1: The same `TypeError` is thrown here and in every other case like this too.

dffml/util/cli/cmds.py
sakshamarora1: Oops! This must've slipped by while copy pasting. Updated! :D

dffml/util/cli/cmd.py
sakshamarora1: We enter a string value in CLI for `-log` which gets changed to it's corresponding int value [here](https://github.com/intel/dffml/blob/master/dffml/util/cli/cmd.py#L26). Providing an int value doesn't work in this case.

pdxjohnny: Sounds good!

dffml/util/cli/cmds.py
pdxjohnny: In `dffml.base:config`,

Before calling

https://github.com/intel/dffml/blob/89401c86f959251f05881a5b23336f1c27712cb0/dffml/base.py#L238

We need to make a `dict`, loop through `cls.__dict__.get('__annotations__', {})`, and check if the key exists in `cls.__dict__`. If it does and it's a `field`, then if it doesn't have a `default` add it to the new `dict`. Then loop through again, do the same thing for fields with a default. Now assign the `cls.__annotations = ` the new `dict`.

dffml/base.py
```diff
@@ -235,7 +259,61 @@ def config(cls):
...
+    defaultdict = {}
+    if len(inspect.getmro(cls)) > 2:
+        class_list = list(inspect.getmro(cls)[:-1])
+    else:
+        class_list = [cls]
+    for clss in class_list:
+        cls_annotations = clss.__dict__.get("__annotations__", {})
+        cls_dict = clss.__dict__
+        for var in cls_annotations:
+            if not dataclasses.is_dataclass(clss):
```
pdxjohnny, line 273: Since the body of the each for loop is the same, we should make the body a function

dffml/service/dev.py
```diff
@@ -60,57 +61,50 @@
...
             if not self.description:
-                self.description = f"DFFML {name} {self.package}"
+                self.description = f"DFFML {self_name} {self.package}"
             # Set target directory to package name if not given
             if not self.target:
                 self.target = self.package
             # Extract
             self.skel.from_template(
-                name,
+                self_name,
```
pdxjohnny, line 107:
```suggestion
                plugin_type,
```

dffml/service/dev.py
```diff
@@ -60,57 +61,50 @@
...
+
+        skel = Skel()
+
+        CONFIG = CreateCMDConfig
 
         async def run(self):
             # Set description if None
             if not self.description:
-                self.description = f"DFFML {name} {self.package}"
+                self.description = f"DFFML {self_name} {self.package}"
```
pdxjohnny, line 101:
```suggestion
                self.description = f"DFFML {plugin_type} {self.package}"
```

dffml/service/dev.py
```diff
@@ -60,57 +61,50 @@
 ]
 
 
-def create_from_skel(name):
+def create_from_skel(self_name):
```
pdxjohnny, line 64: This should probably be something more descriptive (`plugin_type` chosen, since we're creating a package of `plugin_type` from skel/)

```suggestion
def create_from_skel(plugin_type):
```

dffml/util/cli/cmd.py
```diff
@@ -87,21 +89,46 @@ def add_subs(self, add_from: "CMD"):
...
+            if isinstance(arg, Arg):
+                if isinstance(field.metadata.get("position"), int):
+                    position_list[field.metadata.get("position")] = (
+                        field.name,
+                        arg,
+                    )
+                else:
+                    try:
+                        if position_list:
+                            for position in sorted(position_list.keys()):
```
pdxjohnny, line 104: I think you should do the for loop through position_list  after line 112, I suspect this only works if there are two positional arguments that don't have a non-positional one in between them.

dffml/base.py
```diff
@@ -235,7 +259,61 @@ def config(cls):
...
+                    ) or "dataclasses._MISSING_TYPE" not in repr(
+                        clss.__dataclass_fields__.get(var).default_factory
+                    ):
+                        defaultdict[var] = cls_annotations[var]
+
+    defaultdict.update(cls.__annotations__)
+    newdict.update(defaultdict)
+    cls.__annotations__ = newdict
+
+    datacls = dataclasses.dataclass(eq=True, init=False)(cls)
```
pdxjohnny, line 316: I feel like you may have explained, but why was it that `init` must equal `False`?

dffml/util/cli/cmd.py
```diff
@@ -87,21 +89,46 @@ def add_subs(self, add_from: "CMD"):
...
+                                name, positional_arg = position_list[position]
+                                self.add_argument(name, **positional_arg)
+                            position_list.clear()
+                        if not position_list:
+                            self.add_argument("-" + field.name, **arg)
```
pdxjohnny, line 109: We'll want to do `field.name.replace("_", "-")` whenever we add_argument for a field. So that things like, `double_arg` become `-double-arg`

dffml/base.py
sakshamarora1: What happens is that the `__init__` method is not called before [this line](https://github.com/intel/dffml/blob/master/dffml/base.py#L238), so the variables from the base class are not yet in the `__dict__`, so I took the fields from base dataclasses and added them to `__annotations__` of the derived dataclass and changed `init=False` so that same non default argument error would not be thrown and tried other things like adding directly to the `__dict__` of derived class which didn't work either. I guess `init=True` overrides all that extra code.

dffml/service/dev.py
sakshamarora1: Oh, okay!

dffml/util/cli/cmd.py
```diff
@@ -126,6 +128,12 @@ def add_subs(self, add_from: "CMD"):
                 self.add_argument(name.replace("_", "-"), **positional_arg)
             position_list.clear()
 
+        # Add `-log` argument if it's not already added
+        try:
+            self.add_argument(log_cmd.name, **log_cmd)
```
pdxjohnny, line 133: Why the `try` block here?

sakshamarora1: That's for cases when `-log` is already an argument for that command, it throws an `ArgumentError` for conflicting option string in those cases.

pdxjohnny: Ah okay great, let's `except ArgumentError` then :)

dffml/service/dev.py
```diff
@@ -339,25 +352,34 @@ async def run(self):
...
     """
 
-    arg_kwarg = Arg("kwarg", help="Keyword argument to write to stdout")
-    arg_setup_filepath = Arg("setup_filepath", help="Path to setup.py")
+    CONFIG = SetupPyKWArgConfig
 
     @staticmethod
-    def get_kwargs(setup_filepath: str):
-        setup_filepath = Path(setup_filepath)
+    def get_kwargs(setupfilepath: str):
```
sakshamarora1, line 373: `setup_filepath` got converted to `setup-filepath`. Removing the underscore fixed the error for now.

tests/test_cli.py
```diff
@@ -50,15 +50,16 @@ async def setUp(self):
...
         # TODO(p3) For some reason patching Model.load doesn't work
-        # self._stack.enter_context(patch("dffml.model.model.Model.load",
-        #     new=model_load))
         self._stack.enter_context(
-            patch.object(
-                ModelCMD,
-                "arg_model",
-                new=ModelCMD.arg_model.modify(type=model_load),
-            )
+            patch("dffml.model.model.Model.load", new=model_load)
```
sakshamarora1, line 54: I guess this is working now. the `arg_model` method won't work now

service/http/dffml_service_http/cli.py
```diff
@@ -129,78 +154,75 @@ class CreateTLS(TLSCMD):
...
-        default=False,
+    insecure: bool = field(
+        "Start without TLS encryption", action="store_true", default=False,
     )
-    arg_cors_domains = Arg(
-        "-cors-domains",
-        help="Domains to allow CORS for (see keys in defaults dict for aiohttp_cors.setup)",
-        nargs="+",
-        default=[],
+    cors_domains: List[str] = field(
```
sakshamarora1, line 196: The service/http tests are raising `TypeError: '_MISSING_TYPE' object is not iterable` for `cors_domain`. not sure what's happening here. @pdxjohnny Can you provide some input?

sakshamarora1: One of the several places it's raising this: 
```console
ERROR: test_update (tests.test_routes.TestRoutesSource)
  File "/home/tron/Desktop/dffml/dffml/util/asynctestcase.py", line 68, in run_it
  File "/home/tron/anaconda3/lib/python3.7/asyncio/base_events.py", line 587, in run_until_complete
    return future.result()
  File "/home/tron/Desktop/dffml/service/http/tests/test_routes.py", line 349, in setUp
    await super().setUp()
  File "/home/tron/Desktop/dffml/service/http/dffml_service_http/util/testing.py", line 114, in setUp
    await self.tserver.start(self.cli.run())
  File "/home/tron/Desktop/dffml/service/http/dffml_service_http/util/testing.py", line 84, in start
    raise exception
  File "/home/tron/Desktop/dffml/service/http/dffml_service_http/cli.py", line 251, in run
    await self.setup()
  File "/home/tron/Desktop/dffml/service/http/dffml_service_http/routes.py", line 752, in setup
    for domain in self.cors_domains
TypeError: '_MISSING_TYPE' object is not iterable

```

dffml/service/dev.py
pdxjohnny: Sounds good

tests/test_cli.py
```diff
@@ -407,18 +408,18 @@ async def test_record(self):
         subset_urls = list(map(lambda record: record.key, subset))
         results = await Predict.cli(
             "record",
-            "-sources",
-            "primary=json",
-            "-source-filename",
-            self.temp_filename,
+            "-keys",
```
sakshamarora1, line 411: `-source-filename` was not taking anything and was being passed as a flag resulting in the PosixPath('True') error.

A few optional args were changed to positional args so that's why we are facing these errors. Need to find a better approach for positional args I guess.

pdxjohnny: Nice work figuring this out!

dffml/util/cli/cmd.py
```diff
@@ -109,11 +109,9 @@ def add_subs(self, add_from: "CMD"):
         for i, field in enumerate(dataclasses.fields(add_from.CONFIG)):
             arg = mkarg(field)
             if isinstance(arg, Arg):
-                position = field.metadata.get("position", None)
-                if not "default" in arg and position is None:
-                    position = i
-                if isinstance(position, int):
-                    position_list[position] = (field.name, arg)
+                position = None
+                if not "default" in arg:
```
pdxjohnny, line 113: I think this might be what we really want here, because a few of the arguments that you've just defaulted to `None` really are required. I'm sorry I believe I remember saying that we might not have any more cases like this in the meeting, however, `model` and `sources` seem to be those cases where we have a non positional argument that is required.

```suggestion
                if not "default" in arg and not arg.get("required", False):
```

examples/shouldi/shouldi/project/cli.py
```diff
@@ -1,12 +1,34 @@
...
+    )
+    _dbs: List[DependencyDB.load] = field(
+        "Databases to search for info on dependencies",
+        default_factory=lambda: [],
+    )
+    _add: List[DependencyDB.load] = field(
+        "YAML files containing info to supplement or override auto-detected info",
+        default_factory=lambda: [],
+    )
+    _authoritative: List[pathlib.Path] = field(
```
pdxjohnny, line 24: Why are these prefixed with underscores?

sakshamarora1: I prefixed them with underscores to so that they get converted to double dash arguments in cmd.py
I don't know if it's the best solution though. Or we can make them single dash arguments?

pdxjohnny: Let's just make them single dash

PR #580, opened 2020-04-20
Added doctestable examples to db operations

### LINKED ISSUE #576: db: sqlite: Creating lock on __init__ causes issues when not in an event loop

### LINKED ISSUE #577: db: sqlite: remove_query query_values referenced before assigned if no conditions
https://github.com/intel/dffml/blob/89401c86f959251f05881a5b23336f1c27712cb0/dffml/db/sql.py#L214-L247

mHash1m:
Also
Fixes  #576  and
Fixes #577

### DIFF
--- a/dffml/operation/db.py
+++ b/dffml/operation/db.py
@@ -80,6 +80,48 @@ async def db_query(
 async def db_query_create_table(
     self, *, table_name: str, cols: List[str] = []
 ):
+    """
+    Generates a create table query in the database.
+
+    Parameters
+    ++++++++++
+    table_name : str
+        The name of the table to be created.
+    cols : list[str]
+        Columns of the table.
+
+    Examples
+    ++++++++
+
+    >>> sdb = SqliteDatabase(SqliteDatabaseConfig(filename="examples.db"))
+    >>> dataflow = DataFlow(
+    ...     operations={"db_query_create": db_query_create_table.op,},
+    ...     configs={"db_query_create": DatabaseQueryConfig(database=sdb),},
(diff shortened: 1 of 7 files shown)


dffml/operation/db.py
```diff
@@ -80,6 +80,48 @@ async def db_query(
...
+    ++++++++++
+    table_name : str
+        The name of the table to be created.
+    cols : list[str]
+        Columns of the table.
+
+    Examples
+    ++++++++
+
+    >>> sdb = SqliteDatabase(SqliteDatabaseConfig(filename="examples.db"))
```
mHash1m, line 96: left this in example just for create query

dffml/operation/db.py
```diff
@@ -91,6 +133,72 @@ async def db_query_create_table(
...
+    ...         Input(
+    ...             value=[db_query_lookup.op.outputs["lookups"].name],
+    ...             definition=GetSingle.op.inputs["spec"],
+    ...         ),
+    ...     ]
+    ... }
+    >>>
+    >>> async def main():
+    ...     async for ctx, result in MemoryOrchestrator.run(dataflow, inputs):
+    ...         if(result!={}):
```
mHash1m, line 196: The empty brackets in the outputs made it hard to anticipate results (if we print all results, it prints empty dicts for those other than query lookups). Didn't wanna mess with the actual functions just for the sake of examples. I'm open to other suggestions, or if it's a green light I can figure something out inside the functions so they don't return those brackets. Was worried that might break other tests or something.

PR #587, opened 2020-04-28
base: Made classes with default configs to be instantiated easily.

### LINKED ISSUE #522: base: Make it so classes with default configs can be instantiated easily
Right now we have to call `MemoryOrchestrator.withconfig({})`, this is because we still haven't finished converting the `config()` and `args()` calls yet.

But before we do that. We need to modify this

https://github.com/intel/dffml/blob/5d71f28de78224b362644167d04466c15c19e15e/dffml/base.py#L322-L325

to make it so that if `config` is `None` it checks if the `CONFIG` has a default for every field. If it does and `config` is `None`, then we should just create an instance of `CONFIG` and use that. If not, then we should raise the error still.

`dataclasses.fields` will be helpful here.

mHash1m:
Fixes #522 
@pdxjohnny does this seem right? The tests gave no red flags.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Every class and function can now be imported from the top level module
 - `op` attempts to create `Definition`s for each argument if an `inputs` are not
   given.
+- Classes now use `CONFIG` if it has a default for every field and `config` is `None`
 ### Fixed
 - Race condition in `MemoryRedundancyChecker` when more than 4 possible
   parameter sets for an operation.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Operation` inputs and ouputs default to empty `dict` if not given.
 - Ability to export any object with `dffml service dev export`
 - Complete example for dataflow run cli command
+- Tests for default configs instantiation.
 ### Changed
 - `Edit on Github` button now hidden for plugins.
 - Doctests now run via unittests
--- a/dffml/base.py
+++ b/dffml/base.py
@@ -310,17 +310,30 @@ class BaseConfigurableMetaClass(type, abc.ABC):
         def wrapper(self, config: Optional[BaseConfig] = None, **kwargs):
             if config is not None and len(kwargs):
                 raise ConfigAndKWArgsMutuallyExclusive
-            elif config is None and hasattr(self, "CONFIG") and kwargs:
-                try:
-                    config = self.CONFIG(**kwargs)
-                except TypeError as error:
-                    error.args = (
-                        error.args[0].replace(
-                            "__init__", f"{self.CONFIG.__qualname__}"
-                        ),
-                    )
-                    raise
-            if config is None:
+            elif config is None and hasattr(self, "CONFIG"):
+                if kwargs:
+                    try:
+                        config = self.CONFIG(**kwargs)
+                    except TypeError as error:
+                        error.args = (
+                            error.args[0].replace(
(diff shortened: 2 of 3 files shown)


dffml/base.py
```diff
@@ -321,9 +321,17 @@ def wrapper(self, config: Optional[BaseConfig] = None, **kwargs):
                     )
                     raise
             if config is None:
-                raise TypeError(
-                    "__init__() missing 1 required positional argument: 'config'"
-                )
+                use_CONFIG = True
```
pdxjohnny, line 324: Very close! We need to check that it `hasattr(self, "CONFIG")` before using `self.CONFIG`, this and the previous if statement should probably be combined.

So

```python
            elif config is None and hasattr(self, "CONFIG"):
                if kwargs:
                    # old logic
                else:
                    # your new logic
            else:
                raise TypeError(
                    "__init__() missing 1 required positional argument: 'config'"
                )
```

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - `Edit on Github` button now hidden for plugins.
 - Doctests now run via unittests
+- base classes now use `CONFIG` if it has a default for every field and `config` is `None`
```
pdxjohnny, line 20:
```suggestion
- Classes now use `CONFIG` if it has a default for every field and `config` is `None`
```

This is applicable to all classes since you've changed the base class

dffml/base.py
mHash1m: Doesn't this also start throwing the error on other cases not handled previously because of the last `else` clause? Like `if config is not None:` while the kwargs condition may not be fulfilled, it will throw this error. Do we want that?

tests/test_base.py
```diff
@@ -168,3 +169,22 @@ def test_config_set(self):
...
+
+@entrypoint("fake2")
+class FakeTesting2(BaseTesting):
+    
+    CONFIG = FakeTestingConfig2
+
+class TestCONFIG(unittest.TestCase):
+    def test_CONFIG(self):
+        with self.assertRaises(MissingConfig):
+            config = FakeTesting2.config({})
```
pdxjohnny, line 190: We want to test regular instantiation, without the use of the `.config` method or `.withconfig`

```suggestion
            config = FakeTesting2()
```

tests/test_base.py
```diff
@@ -168,3 +169,22 @@ def test_config_set(self):
...
+    label: str = "unlabeled"
+
+@entrypoint("fake2")
+class FakeTesting2(BaseTesting):
+    
+    CONFIG = FakeTestingConfig2
+
+class TestCONFIG(unittest.TestCase):
+    def test_CONFIG(self):
+        with self.assertRaises(MissingConfig):
```
pdxjohnny, line 189:
```suggestion
        with self.assertRaisesRegex(TypeError, "__init__() missing 1 required positional argument: 'config'"):
```

mHash1m: I just did this because it threw that error at me :D Not getting the TypeError yet. Gonna try regular instantiation.

mHash1m: @pdxjohnny Also how do I set the default value right after this test to test the successful case?

pdxjohnny: Probably best to just make another class 

```python
@config
class FakeTestingConfig3:
    label: str = "unlabeled"

@entrypoint("fake3")
class FakeTesting3(BaseTesting):

    CONFIG = FakeTestingConfig3
```

mHash1m: Yeah crossed my mind, was reluctant. Thanks!
`with self.assertRaises(TypeError):` working with the normal instantiation now. `self.assertRaisesRegex` throws assertion error. 
Pushing in a bit.

pdxjohnny:
Great work! Very happy to see this merged :)

PR #594, opened 2020-04-29
df:base: create Definition for each args

### LINKED ISSUE #189: df: base: op should extrapolate inputs based off typing information
Related: #188

```python
    >>> @op(inputs={"host": HOST, "port": PORT, "message": MESSAGE})
    ... async def send_to_server(host: str, port: int, message: str):
    ...     reader, writer = await asyncio.open_connection(host, port)
    ...
    ...     writer.write(message.encode())
    ...     await writer.drain()
    ...
    ...     data = await reader.read(100)
    ...     print(f"Client sent {message!r}, got: {data.decode()!r}")
    ...
    ...     writer.close()
    ...     await writer.wait_closed()
```

We had to pass `inputs` to op and create all the `Defintion` instances.

What would be good is to modify `dffml/df/base.py` so that if `op` is not provided with `inputs` it uses `inspect.signature` to look at the functions arguments. It should create `Defintion`s for each argument using the `str(of the data type)` as the primitive and the name as the name.

Also add some validation to make sure that we can only do this for data types which are already primitives, aka, int, float, str, etc.

sk-ip:
closes #189 

Signed-off-by: Sudhanshu Kumar <Sudhanshukumar5459@gmail.com>

### DIFF
--- a/dffml/df/base.py
+++ b/dffml/df/base.py
@@ -193,6 +193,20 @@ def op(imp_enter=None, ctx_enter=None, config_cls=None, **kwargs):
         if not "conditions" in kwargs:
             kwargs["conditions"] = []
 
+        primitive_types = ["int", "float", "str", "bool"]
+
+        if not "inputs" in kwargs:
+            sig = str(inspect.signature(func))
+            sig = [
+                arg.strip()
+                for arg in sig.replace("(", "").replace("(", "").split(",")
+            ]
+
+            for arguements in sig:
+                arg, argtype = arguements.split(":")
+                if argtype in primitive_types:
+                    Definition(name=arg, primitive=str(argtype))
+
         func.op = Operation(**kwargs)
         func.ENTRY_POINT_NAME = ["operation"]
         cls_name = (
--- a/dffml/df/base.py
+++ b/dffml/df/base.py
@@ -193,19 +193,17 @@ def op(imp_enter=None, ctx_enter=None, config_cls=None, **kwargs):
         if not "conditions" in kwargs:
             kwargs["conditions"] = []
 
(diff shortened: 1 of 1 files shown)


sk-ip:
@pdxjohnny , the Definition is left hanging and not assigned anywhere, where should I assign the Definition instances.

dffml/df/base.py
```diff
@@ -193,6 +193,20 @@ def wrap(func):
         if not "conditions" in kwargs:
             kwargs["conditions"] = []
 
+        primitive_types = ["int", "float", "str", "bool"]
+
+        if not "inputs" in kwargs:
+            sig = str(inspect.signature(func))
```
pdxjohnny, line 199: Let's not parse the string, look on line [222](https://github.com/intel/dffml/pull/594/files#diff-9433e166ad553c4951da0eb1beb170c8R222) to see how to get each parameter.

dffml/df/base.py
```diff
@@ -193,6 +193,20 @@ def wrap(func):
...
+            sig = [
+                arg.strip()
+                for arg in sig.replace("(", "").replace("(", "").split(",")
+            ]
+
+            for arguements in sig:
+                arg, argtype = arguements.split(":")
+                if argtype in primitive_types:
+                    Definition(name=arg, primitive=str(argtype))
```
pdxjohnny, line 208: We should make `kwargs["inputs"]` a empty dict.

Each key should be the name of the parameter, the value should be this newly created `Definition`

dffml/df/base.py
sk-ip: Done

pdxjohnny:
Great work @sk-ip! :D

PR #609, opened 2020-05-03
model: make model plugins import third party libraries dynamically.

### LINKED ISSUE #320: model: Dynamiclly import libraries for model plugins when plugin is instantiated
Importing libraries takes time. If we were to delay loading of libraries that various plugins require until the classes for those plugins are instantiated, listing of available plugins, and other things would be much faster.

To complete this issue, change the models in `model/` to import the third party libraries they require (aka `numpy`, `sklearn`, `tensorflow`, etc.) via `importlib`. Store the references to the imported libraries as properties in the instantiated classes. (for instance, `self._tensorflow = importlib.import_module("tensorflow")`

mHash1m:
Resolves #320 
I left out a few sub-modules because I wasn't sure if they should also be dynamically imported. Please let me know if I should also push that along this PR, if you see fit.

### DIFF
--- a/model/scikit/dffml_model_scikit/scikit_base.py
+++ b/model/scikit/dffml_model_scikit/scikit_base.py
@@ -6,11 +6,10 @@ Base class for Scikit models
 import json
 import hashlib
 import pathlib
+import importlib
+
 from typing import AsyncIterator, Tuple, Any, NamedTuple
 
-import joblib
-import numpy as np
-import pandas as pd
 from sklearn.metrics import silhouette_score, mutual_info_score
 
 from dffml.record import Record
@@ -30,6 +29,9 @@ class ScikitConfig(ModelConfig, NamedTuple):
 class ScikitContext(ModelContext):
     def __init__(self, parent):
         super().__init__(parent)
+        self.pd = importlib.import_module("pandas")
+        self.np = importlib.import_module("numpy")
+        self.joblib = importlib.import_module("joblib")
         self.features = self.applicable_features(self.parent.config.features)
         self._features_hash = self._feature_predict_hash()
         self.clf = None
@@ -60,7 +62,7 @@ class ScikitContext(ModelContext):
 
     async def __aenter__(self):
         if self._filepath.is_file():
-            self.clf = joblib.load(str(self._filepath))
+            self.clf = self.joblib.load(str(self._filepath))
         else:
             config = self.parent.config._asdict()
             del config["directory"]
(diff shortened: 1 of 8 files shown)


model/scikit/dffml_model_scikit/scikit_base.py
```diff
@@ -30,6 +27,10 @@ class ScikitConfig(ModelConfig, NamedTuple):
 class ScikitContext(ModelContext):
     def __init__(self, parent):
         super().__init__(parent)
+        self.pd = importlib.import_module("pandas")
+        self.np = importlib.import_module("numpy")
+        self.joblib = importlib.import_module("joblib")
+        self.hashlib = importlib.import_module("hashlib")
```
pdxjohnny, line 33: We only want to import third party libraries dynamically, so hashlib should stay as a regular `import`

model/scikit/dffml_model_scikit/scikit_base.py
```diff
@@ -240,29 +241,31 @@ class Scikit(Model):
     def __init__(self, config) -> None:
         super().__init__(config)
         self.saved = {}
+        self.hashlib = importlib.import_module("hashlib")
+        self.json = importlib.import_module("json")
```
pdxjohnny, line 245: Same with `json`, regular `import` statement for modules that are in the standard library

model/tensorflow/dffml_model_tensorflow/dnnc.py
```diff
@@ -4,16 +4,15 @@
 """
 import os
 import abc
-import hashlib
-import inspect
```
pdxjohnny, line 8: Same thing, inspect and hashlib are in the stdlib

model/tensorflow_hub/dffml_model_tensorflow_hub/text_classifier.py
```diff
@@ -5,16 +5,13 @@
 """
 # TODO Add docstrings
 import os
-import hashlib
```
pdxjohnny, line 8: Same thing, hashlib is in the stdlib

model/transformers/dffml_model_transformers/ner/ner_model.py
```diff
@@ -168,38 +162,51 @@ class NERModelConfig:
...
     def __init__(self, parent, **kwconfig):
         super().__init__(parent)
+        self.pd = importlib.import_module("pandas")
+        self.np = importlib.import_module("numpy")
+        self.tf = importlib.import_module("tensorflow")
+        self.re = importlib.import_module("re")
+        self.math = importlib.import_module("math")
+        self.glob = importlib.import_module("glob")
+        self.datetime = importlib.import_module("datetime")
+        self.collections = importlib.import_module("collections")
```
pdxjohnny, line 209: Same thing, re, math, glob, datetime, and collections, are in the stdlib. Also, we can replace glob with `pathlib(the_path).glob(arguments)`

model/scikit/dffml_model_scikit/scikit_base.py
mHash1m: Ah! My bad, I misinterpreted third-party modules as those other than dffml itself. I'll fix it soon.

PR #613, opened 2020-05-05
cli:list: change print to yield

### LINKED ISSUE #610: cli: list: Change to json dump
Listing records shows the following (dffml list records ...):

```
$ cat > data.csv << EOF                                           
key,third_party,maintained
dffml,True,True
python,False,True
EOF
$ dffml list records -sources data=csv -source-filename data.csv
dffml
third_party                   True
maintained                    True
Predictions
Undetermined
python
third_party                   False
maintained                    True
Predictions
Undetermined
```

We should make it look better somehow, or just yield each one and not print so it gets dumped as json

sk-ip:
closes #610

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   given.
 - Classes now use `CONFIG` if it has a default for every field and `config` is `None`
 - Models now dynamically import third party modules.
-- In CLI, Records from Sources gets yielded instead of printing it.
+- `dffml list records` command prints Records as JSON using `.export()`
 ### Fixed
 - Race condition in `MemoryRedundancyChecker` when more than 4 possible
   parameter sets for an operation.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   given.
 - Classes now use `CONFIG` if it has a default for every field and `config` is `None`
 - Models now dynamically import third party modules.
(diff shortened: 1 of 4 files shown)


sk-ip:
Wow,i will fix the tests.....

pdxjohnny:
@sk-ip since you're yielding, `.cli` will return a list of record objects, 

```python
            records = await CLI.cli(
                "list",
                "records",
                "-sources",
                "tmp=json",
                "-source-tmp-filename",
                filename,
            )
        # See below for demo of this
        records = dict(map(lambda r: (r["key"], r), records))
        for key in keys:
            self.assertIn(key, records)
```

```console
$ python
Python 3.7.7 (default, Apr 14 2020, 15:30:21)
[GCC 7.5.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> records = [{"key": "a"}, {"key": "b"}]
>>> records = dict(map(lambda r: (r["key"], r), records))
>>> records
{'a': {'key': 'a'}, 'b': {'key': 'b'}}
```

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   given.
 - Classes now use `CONFIG` if it has a default for every field and `config` is `None`
 - Models now dynamically import third party modules.
+- In CLI, Records from Sources gets yielded instead of printing it.
```
pdxjohnny, line 31:
```suggestion
- `dffml list records` command prints Records as JSON using `.export()`
```

pdxjohnny:
Sweet! Thank you! (I just wanted to tweak the changelog a bit so that the description was more of what the users will see differently, rather than what happened in the code)

Issue #565, opened 2020-04-14
docs: conf: Replace add_javascript with add_js_file


pdxjohnny:
```
2020-04-14T21:52:55.5827167Z /home/runner/work/dffml/dffml/docs/conf.py:92: RemovedInSphinx40Warning: The app.add_javascript() is deprecated. Please use app.add_js_file() instead.
```

shivam-bit:
Hi, I would like to fix this issue.

shivam-bit:
@pdxjohnny issue #565 fixed

PR #626, opened 2020-05-10
Updated changelog with issue #565 fixes

shivam-bit:
Updated changelog with issue #565 fixes


pdxjohnny:
It's okay since that change isn't hardly user facing at all we don't need a changelog entry for it. Thank you for fixing this!

PR #627, opened 2020-05-11
df: base: create Definition for spec

### LINKED ISSUE #601: df: base: Auto create definitions when they might have a spec
Related: #594, #598

sk-ip:
the Definition for spec (NamedTuple) types will be created.

fixes #601 


dffml/df/base.py
```diff
@@ -216,8 +216,10 @@ def wrap(func):
                         ),
                     )
                 else:
```
pdxjohnny, line 218: This should be in an `elif` where we check if the definition is a NamedTuple, or if it's a dataclass, if it is either, then we do this new code you've added. Otherwise we keep the old else statement.

be careful though because you can't just do `issubclass(..., NamedTuple)`,

```python
>>> from typing import NamedTuple
>>> import dataclasses
>>>
>>> class SubN(NamedTuple):
...   i: int
...
>>> @dataclasses.dataclass
... class SubD:
...   i: int
...
>>> print(bool(issubclass(SubN, NamedTuple)))
False
>>> print(dataclasses.is_dataclass(SubD))
True
>>> print(bool(issubclass(SubN, tuple) and hasattr(SubN, "_asdict")))
True
>>>
```

sk-ip: done, thanks

dffml/df/base.py
```diff
@@ -215,6 +216,15 @@ def wrap(func):
...
                     )
+                elif is_dataclass(param.annotation) or bool(
+                    issubclass(param.annotation, tuple)
+                    and hasattr(param.annotation, "_asdict")
+                ):
+                    kwargs["inputs"][name] = Definition(
+                        name=".".join(name_list),
+                        primitive="map",
+                        spec=param.annotation,
+                    )
```
pdxjohnny, line 227: When you implment the same thing but subspec, you'll have an if statement after the one you added where you check if the `__origin__` (use typing.get_origin, also in base.py but you would probably move it to util.data). If origin is Typing.list then set primitive to array, otherwise set it to map if dict. If it's neither of those, then fall through to the last case.

You'll still want to have the logic you added in this if statement to validate that the list / dict has a inner type which is a Named Tuple or dataclass

pdxjohnny: You'll also want to modify the code in Input.__init__ so that it checks what it's primitive is, and throws an error if the datatype passed isn't right type, so if it's not a list or dict, rather than the current behavior which just checks if the value is the correct.

sk-ip:
Adding a checklist for myself

- [ ] Complete the subspec part
- [ ] Write tests for spec / subspec 
- [ ] Add tests in the examples of the docstrings in `@op`
- [ ] Add extra validation to the `Input.__init__` side

dffml/df/base.py
sk-ip: thanks, done

dffml/df/base.py
```diff
@@ -215,6 +217,38 @@ def wrap(func):
...
+                    if is_dataclass(innerclass) or bool(
+                        issubclass(innerclass, tuple)
+                        and hasattr(innerclass, "_asdict")
+                    ):
+                        kwargs["inputs"][name] = Definition(
+                            name=".".join(name_list),
+                            primitive=primitive,
+                            spec=innerclass,
+                            subspec=True,
+                        )
```
pdxjohnny, line 251: Very nice! Can we add unit tests for `op` to check that this is working as expected?

pdxjohnny: Or even just add example to the docstring of `op`, they'll get tested, that'd be even better

pdxjohnny: The docstring of `Definition` might be helpful: https://intel.github.io/dffml/master/api/df/types.html#dffml.df.types.Definition

sk-ip: Added the docstring examples, thanks

dffml/df/base.py
```diff
@@ -184,6 +192,45 @@ def op(*args, imp_enter=None, ctx_enter=None, config_cls=None, **kwargs):
...
+
+    >>> from dffml import Definition, Input, op
+    >>> from typing import NamedTuple, List, Dict
+    >>>
+    >>> class Person(NamedTuple):
+    ...     name: str
+    ...     age: int
+    ...
+    >>> @op
+    ... def isTeen(p: List[Person]):
```
pdxjohnny, line 207: Let's just make this all about voting

```suggestion
    ... def cannotVote(p: List[Person]):
```

dffml/df/base.py
```diff
@@ -184,6 +192,45 @@ def op(*args, imp_enter=None, ctx_enter=None, config_cls=None, **kwargs):
...
+    ...     return [True if person.age > 18 else False for person in p]
```
pdxjohnny, line 208:
```suggestion
    ...     return list(filter(lambda person: person.age < 18, p))
```

The `>` here should have been `<`

dffml/df/base.py
```diff
@@ -184,6 +192,45 @@ def op(*args, imp_enter=None, ctx_enter=None, config_cls=None, **kwargs):
...
+    >>> Input(
+    ...     value=[
+    ...         {"name": "Bob", "age": 20},
+    ...         {"name": "Mark", "age": 21},
+    ...         {"name": "Alice", "age": 90},
+    ...     ],
+    ...     definition=isTeen.op.inputs["p"],
+    ... )
+    Input(value=[Person(name='Bob', age=20), Person(name='Mark', age=21), Person(name='Alice', age=90)], definition=isTeen.p)
+    >>> @op
```
pdxjohnny, line 220: Blank line for clarity / consistency

```suggestion
    >>>
    >>> @op
```

dffml/df/base.py
```diff
@@ -184,6 +192,45 @@ def op(*args, imp_enter=None, ctx_enter=None, config_cls=None, **kwargs):
...
+    ... def canVote(p: Dict[str, Person]):
+    ...     return [True if person.age > 18 else False for name,person in p.items()]
```
pdxjohnny, line 222: Black formatting and consistency with above

```suggestion
    ...     return {
    ...         person.name: person
    ...         for person in filter(lambda person: person.age >= 18, p.values())
    ...     }
```

dffml/df/base.py
```diff
@@ -184,6 +192,45 @@ def op(*args, imp_enter=None, ctx_enter=None, config_cls=None, **kwargs):
...
+    >>>
+    >>> Input(
+    ...     value=[
+    ...         {"name": "Bob", "age": 20},
+    ...         {"name": "Mark", "age": 21},
+    ...         {"name": "Alice", "age": 90},
+    ...     ],
+    ...     definition=isTeen.op.inputs["p"],
+    ... )
+    Input(value=[Person(name='Bob', age=20), Person(name='Mark', age=21), Person(name='Alice', age=90)], definition=isTeen.p)
```
pdxjohnny, line 219:
```suggestion
    Input(value=[Person(name='Bob', age=20), Person(name='Mark', age=21), Person(name='Alice', age=90)], definition=cannotVote.p)
```

dffml/df/base.py
```diff
@@ -184,6 +192,45 @@ def op(*args, imp_enter=None, ctx_enter=None, config_cls=None, **kwargs):
...
+    ...     return [True if person.age > 18 else False for person in p]
+    ...
+    >>>
+    >>> Input(
+    ...     value=[
+    ...         {"name": "Bob", "age": 20},
+    ...         {"name": "Mark", "age": 21},
+    ...         {"name": "Alice", "age": 90},
+    ...     ],
+    ...     definition=isTeen.op.inputs["p"],
```
pdxjohnny, line 217:
```suggestion
    ...     definition=cannotVote.op.inputs["p"],
```

PR #668, opened 2020-05-22
noasync: expose run from high level to noasync

### LINKED ISSUE #665: noasync: Expose high level run
We need the `noasync` API to also include the  `run` function within `dffml.high_level`.

When you get to the docstring portion of things be aware that the docstring for `run` will probably need to be re-written because of `send_to_server` won't still work with the simple find and replace that is being done on the other docstrings

You'll probably just need to replace it with some regular socket calls

sk-ip:
closes #665 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   path to directory or Git repo URL and runs the appropriate static analyzers.
 - Support for entrypoint style loading of operations and seed inputs in `dataflow create`.
 - Definition for output of the function that `op` wraps.
-- Expose high level load, run and save funtions to noasync. 
+- Expose high level load, run and save funtions to noasync.
 ### Changed
 - `edit` command substituted with `edit record`
 - `Edit on Github` button now hidden for plugins.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   path to directory or Git repo URL and runs the appropriate static analyzers.
 - Support for entrypoint style loading of operations and seed inputs in `dataflow create`.
 - Definition for output of the function that `op` wraps.
-- Expose high level load, run and save funtions to noasync.
+- Expose high level load, run and save functions to noasync.
 ### Changed
 - `edit` command substituted with `edit record`
 - `Edit on Github` button now hidden for plugins.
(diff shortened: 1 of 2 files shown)


dffml/noasync.py
```diff
@@ -114,3 +114,90 @@ def cleanup():
...
     )
 )
+
+
+def run(*args, **kwargs):
+    """
+    >>> # Socket server derived from
+    >>> # https://docs.python.org/3/library/socketserver.html#asynchronous-mixins
+    >>> import socket
+    >>> import threading
```
pdxjohnny, line 124:
```suggestion
```

dffml/noasync.py
```diff
@@ -114,3 +114,90 @@ def cleanup():
...
+    >>> import socketserver
+    >>>
+    >>> from dffml.noasync import run
+    >>> from dffml import DataFlow, Input, op
+    >>>
+    >>> class ThreadedTCPRequestHandler(socketserver.BaseRequestHandler):
+    ...     def handle(self):
+    ...         data = str(self.request.recv(1024), "ascii")
+    ...         cur_thread = threading.current_thread()
+    ...         response = bytes("{}: {}".format(cur_thread.name, data), "ascii")
```
pdxjohnny, line 134: Let's remove the use of the `threading`, we don't use the threads name, in fact, you remove it below. So let's just not add it in the first place.

dffml/noasync.py
sk-ip: maybe, we need this.

dffml/noasync.py
sk-ip: yes is removed it from the below, because if we run this code on different machines the thread names will come differently, hence the tests will fail sometimes.

CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   path to directory or Git repo URL and runs the appropriate static analyzers.
 - Support for entrypoint style loading of operations and seed inputs in `dataflow create`.
 - Definition for output of the function that `op` wraps.
+- Expose high level load, run and save funtions to noasync.
```
pdxjohnny:
```suggestion
- Expose high level load, run and save functions from noasync
```

pdxjohnny: Misspelling of functions ;)

dffml/noasync.py
pdxjohnny: You are quite right! I'm sorry

sk-ip:
thanks for pointing that out :)

pdxjohnny:
Sweet! Thank you! Merged :D

PR #679, opened 2020-05-26
Stop guessing directory for models

### LINKED ISSUE #655: model: Stop guessing directory for models
We've realized that the hashing of properties in the config hides things from
users and makes using DFFML confusing.

We should make the directory property required so that users always know where
their model are being stored, and they won't have to re-specify parameters. We
may also need to save the original config in case we need to re-build model
parameters when they aren't passed in the same way for accuracy / predict
commands.

sudharsana-kjl:
fix #655


dffml/model/model.py
```diff
@@ -77,19 +77,22 @@ class Model(BaseDataFlowFacilitatorObject):
     """
 
     CONFIG = ModelConfig
+    DIRECTORY = None
```
pdxjohnny, line 80:
```suggestion
```

pdxjohnny:
https://github.com/intel/dffml/blob/3152df7bd3731c33602ea1c21efed20f180b9017/model/transformers/dffml_model_transformers/ner/ner_model.py#L48-L53

pdxjohnny:
https://github.com/intel/dffml/blob/3152df7bd3731c33602ea1c21efed20f180b9017/model/transformers/dffml_model_transformers/ner/ner_model.py#L62-L65

pdxjohnny:
Nice!!! Thank you!!!

PR #684, opened 2020-05-28
WIP: plugin: Plugin for secrets storage

aghinsa:


dffml/secrets/from_file.py
```diff
@@ -0,0 +1,51 @@
...
+
+from ..base import config
+from ..util.entrypoint import entrypoint
+from .base import BaseSecretsContext,BaseSecrets
+
+@config
+class FileSecretsConfig:
+    filename: str
+
+class FileSecretsContext(BaseSecretsContext):
```
pdxjohnny, line 15: I would say you should probably just subclass from `INISourceContext`

dffml/secrets/from_file.py
```diff
@@ -0,0 +1,51 @@
...
+        self.fd.close()
+        await super().__aexit__(*args,**kwargs)
+
+@entrypoint("file")
+class FileSecrets(BaseSecrets, BaseSecretsContext):
+
+    CONTEXT = FileSecretsContext
+    CONFIG = FileSecretsConfig
+
+    def __init__(self, cfg):
```
pdxjohnny, line 49: Unneeded

dffml/secrets/from_file.py
pdxjohnny: Sorry, this should be a subclass of `MemorySourceContext`, and `FileSecrets` should be a subclass of `INISource`, that way you can define `set` and `get` methods which just grab from `self.mem`. I'd recommend splitting on `.` and using the zeroith index as the key into `self.mem` if the split results in a single element list, then use a key of `__global__` or something.

dffml/secret/file.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+
+class FileSecretContext(BaseSecretContext, MemorySourceContext):
+    def __init__(self,*args,**kwargs):
+        super().__init__(*args,**kwargs)
+
+
+    async def get(self, name: str) -> Union[None, str]:
+        with open(self.parent.config.filename,'r+') as fd:
```
pdxjohnny, line 23: It looks like we shouldn't need this, because `FileSource` calls `load_fd` of `INISource` on `__aenter__`, which will load the data into `self.mem`

dffml/secret/file.py
```diff
@@ -0,0 +1,46 @@
...
+
+@entrypoint("file")
+class FileSecret(BaseSecret,INISource):
+
+    CONTEXT = FileSecretContext
+    CONFIG = INISourceConfig
+
+    def __init__(self,*args,**kwargs):
+        super().__init__(*args,**kwargs)
+        self.mem: Dict[str, Record] = {}
```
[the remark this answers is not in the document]
pdxjohnny: We should also probably issue a warning about this being insecure in this type of style (with the extra newlines so people are sure to see it)

```python
        warnings.warn(
            """


            INSECURE This really does pbkdf2_hmac and not scrypt (since that
            requires openssl 1.1 and most systems only have 1.0.x at time of
            writing) Also its just the example I copy pasted from the docs to
            illustrate threading. 100000 is probably not enough iterations!!!

            """
        )
```

dffml/secret/file.py
```diff
@@ -0,0 +1,46 @@
...
+
+class FileSecretContext(BaseSecretContext, MemorySourceContext):
+    def __init__(self,*args,**kwargs):
+        super().__init__(*args,**kwargs)
+
+
+    async def get(self, name: str) -> Union[None, str]:
+        with open(self.parent.config.filename,'r+') as fd:
+            await self.parent.load_fd(fd)
+        record = self.parent.mem.get(name,None)
```
pdxjohnny, line 25: I'm also realizing that this will end up with `.ini` files that look like:

```ini
[thing0]
data = data0

[thing1]
data = data1

[thing2]
data = data2
```

Whereas we would probably be better off doing:

```ini
[secrets]
thing0 = data0
thing1 = data1
thing2 = data2
```

dffml/secret/file.py
```diff
@@ -0,0 +1,46 @@
...
+    async def set(self, name: str, value: Any):
+        record = Record(name, data={"features": {"data": value}})
+        await self.update(record)
+        with open(self.parent.config.filename,'w+') as fd:
+            await self.parent.dump_fd(fd)
+
+
+
+@entrypoint("file")
+class FileSecret(BaseSecret,INISource):
```
pdxjohnny, line 39: I'm also realizing, we should probably call this `INISecret`, sorry for the confusion

dffml/secret/file.py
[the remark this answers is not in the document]
aghinsa: > We should probably put mem in the `Context` since the parent isn't going to use it

We are using `update` and `load_fd` which expects `parent` to have `mem`. Am I missing something here?

dffml/secret/ini.py
```diff
@@ -0,0 +1,47 @@
+import warnings
+from typing import Union, Any
+
+from ..base import config
+from dffml.high_level import Record
```
pdxjohnny, line 5:
```suggestion
from ..record import Record
```

operations/deploy/dffml_operations_deploy/operations.py
```diff
@@ -27,6 +27,7 @@ def get_url_from_payload(payload: Dict[str, Any]):
     outputs={"is_default_branch": is_default_branch},
 )
 def check_if_default_branch(payload):
+    print(f"Payload received : {payload.keys()}")
```
pdxjohnny, line 30:
```suggestion
```

service/http/dffml_service_http/routes.py
```diff
@@ -246,17 +246,18 @@ async def multicomm_dataflow(self, config, request):
                         status=HTTPStatus.NOT_FOUND,
                     )
                 if preprocess_mode == "json":
+                    print(f"Header are : {request.headers}")
```
pdxjohnny, line 249:
```suggestion
```

PR #685, opened 2020-05-29
docs: model: vowpalWabbit: added example usage

### LINKED ISSUE #615: docs: model: vowpal: Missing example usage
Ping @0dust

0dust:
Fixes #615 
Cleanup and Support for data already in vw format.


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+- Vowpal Wabbit example usage
```
pdxjohnny, line 9:
```suggestion
```

model/vowpalWabbit/dffml_model_vowpalWabbit/vw_base.py
```diff
@@ -373,6 +394,57 @@ async def predict(
...
+
+    Output
+
+    .. code-block:: json
+
+        [
+            {
+                "extra": {},
+                "features": {
+                    "A": "| price:.46 sqft:.4 age:.10 1924"
```
pdxjohnny, line 433: What's up with this?

CHANGELOG.md
pdxjohnny: Since we've already said we add vowpal wabbit this release

pdxjohnny:
@0dust just pinging you on this in case it dropped off your radar

model/vowpalWabbit/dffml_model_vowpalWabbit/vw_base.py
0dust: This is the standard input format taken by vowpal wabbit. So a feature can be a string in the specific format or a bunch of columns that get converted to this format if `convert_to_vw` flag is true.

pdxjohnny: This shouldn't be the way we store that within a record though. It looks to me like this should be

```python
                "features": {
                    "price": 46,
                    "sqft": 4,
                    "age":10,
                }
```

Then it's unclear to me what `1924` is all about. The point here is that the features should be stored in the standard format when they're in the `Record` and if `convert_to_vw` flag is needed to change them into the vw format, then that should probably default to True.

I feel like I may be missing something here though?

model/vowpalWabbit/dffml_model_vowpalWabbit/vw_base.py
```diff
@@ -77,7 +70,7 @@ class VWConfig:
     )
     convert_to_vw: bool = field(
```
pdxjohnny, line 71: Let's change the name so that it's a default of False, because when we give it on the command line a bool flag that has a default of False, to set it to true all we need to do is say `-flag` rather than `-flag false` or `-flag true`

model/vowpalWabbit/dffml_model_vowpalWabbit/vw_base.py
```diff
@@ -253,7 +246,7 @@ async def train(self, sources: Sources):
             )
             data.append(feature_data)
         vw_data = pd.DataFrame(data)
-        if self.parent.config.convert_to_vw:
+        if self.parent.config.requires_formatting:
```
pdxjohnny, line 249: Following the above example

```suggestion
        if not self.parent.config.noconvert:
```

model/vowpalWabbit/dffml_model_vowpalWabbit/vw_base.py
```diff
@@ -76,7 +69,7 @@ class VWConfig:
         "Feature to be used as `tag` in conversion of data to vowpal wabbit input format",
         default=None,
     )
-    convert_to_vw: bool = field(
+    requires_formatting: bool = field(
```
pdxjohnny, line 72: I'm sorry, but we've gone back to the original behavior here.

From the [meeting minutes](https://docs.google.com/document/d/16u9Tev3O0CcUDe2nfikHmrO3Xnd4ASJ45myFgQLpvzM/edit#heading=h.hmuqg6k5xmp4):

> Because we want to be able to loop through all models with minimal tweaks to configs, let’s change conver_to_vw() to something a flag that indicates feature data is already in vw format, so that users opt-out of covert_to_vw() rather than opt-in.

For example

```python
    noconvert: bool = field(
        "Do not convert record features to vowpal wabbit input format",
        default=False,
    )

File: model/vowpalWabbit/examples/accuracy.sh [hunk 6]
```diff
@@ -0,0 +1,9 @@
+dffml accuracy  \
+  -model vwmodel \
```
pdxjohnny, line 2: Following the above example

```suggestion
  -model vwmodel \
  -model-noconvert \
```

pdxjohnny:
Very nice!! :D

PR #694, opened 2020-06-05
examples: ffmpeg : Use entrypoint loading

### LINKED ISSUE #631: operations: deploy: Support secret for webhook
We need to open an issue to make sure that we support having this only happen
with a secret.

We'll need to add some operation that verifies that secret == what's supposed to
be. and we'll set it via a `@config` for the operation. And then make it output
a bool which we can use as a condition.

### LINKED ISSUE #658: examples: ffmpeg: Transform from package into single Python file
Now that we can do entrypoint style loading of operations we should change the
Continuous Deployment of DataFlows example so that the ffmpeg package is really
just a single Python file with the operation and dataflow in it.

aghinsa:
fixes #631 
fixes #658


docs/usage/webhook/webhook.rst
```diff
@@ -32,7 +33,14 @@ Setup a http server in ``ffmpeg/deploy/webhook``, to receive webhook and redploy
     restart_running_containers
     cleanup_git_repo
     EOF
-    $ dffml dataflow create -config yaml $(cat /tmp/operations) > deploy/webhook/df/webhook.yaml
+    $ dffml dataflow create -config yaml $(cat /tmp/operations) \
+        -seed './deploy/webhook/secret.ini'=ini_file > deploy/webhook/df/webhook.yaml
```
pdxjohnny, line 37:
```suggestion
        -config \
            ini=check_secret_match.secret.plugin \
            './deploy/webhook/secret.ini'=check_secret_match.secret.config.filename \
        > deploy/webhook/df/webhook.yaml
```


```yaml
config:
    check_secret_match:
        secret:
            plugin: ini
            config:
                 filename: ./deploy/webhook/secret.ini
```

PR #696, opened 2020-06-07
cli: dataflow: Add flow to dataflow create

### LINKED ISSUE #661: cli: dataflow: create: Modify InputFlow of operations
When we create a DataFlow we should have some syntax that allows us to modify
the `InputFlow` of operations within the `DataFlow.flow` dictionary.

We can probably leverage `dffml.util.data:traverse_get` (and might need to make
`traverse_set`) as well as the `dffml.util.cli.parser:ParseInputsAction`

aghinsa:
fixes #661

### DIFF
--- a/dffml/cli/dataflow.py
+++ b/dffml/cli/dataflow.py
@@ -17,7 +17,7 @@ from ..configloader.json import JSONConfigLoader
 from ..source.source import SubsetSources, Sources
 from ..source.json import JSONSource
 from ..source.file import FileSourceConfig
-from ..util.data import merge, split_dot_seperated
+from ..util.data import merge, split_dot_seperated, traverse_set
 from ..util.entrypoint import load
 from ..util.cli.cmd import CMD, CMDOutputOverride
 from ..util.cli.cmds import (
@@ -81,6 +81,9 @@ class CreateConfig:
     flow: List[str] = field(
         "Flow of inputs", action=ParseInputsAction, default_factory=lambda: [],
     )
+    config: List[str] = field(
+        "configs", action=ParseInputsAction, default_factory=lambda: [],
+    )
 
 
 class Create(CMD):
@@ -116,7 +119,8 @@ class Create(CMD):
                         dataflow.flow[opname].inputs[key] = v
                     else:
                         dataflow.flow[opname].conditions = v
-
+                for v, k in self.config:
+                    traverse_set(dataflow.configs, k, value=v)
                 exported = dataflow.export(linked=not self.not_linked)
                 print((await loader.dumpb(exported)).decode())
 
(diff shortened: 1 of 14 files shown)


dffml/cli/dataflow.py
```diff
@@ -104,6 +107,15 @@ async def run(self):
                     for val, def_name in self.seed
                 ]
                 dataflow.seed.extend(self.seed)
+
+                # flow argument key is of the form opname.inputs/conditions.keyname
+                for v, k in self.flow:
+                    opname, val_type, key = k.split(".")
```
pdxjohnny, line 113: We need to handle the case where there are more than three `.`, possible solution is to split, use last two, join rest of leading to use as first one.

pdxjohnny: Let's use `split_dot_separated` here.

pdxjohnny: Let's add support for setting config via:

```python
                for v, k in self.config:
                    k = split_dot_separated (k)
                    traverse_set(dataflow.config, k, v)
```

dffml/util/data.py
```diff
@@ -109,13 +131,45 @@ def traverse_get(target, *args):
...
+    """
+    if len(args) == 1:
+        args = split_dot_seperated(args[0])
+    if len(args) == 1:
+        target[args[0]] = value
+        return
+
+    current = target
+    for level in args[:-1]:
+        current = current[level]
```
pdxjohnny, line 167: This might not be correct, but the point it we need to be able to create dict's if they don't exit on the way down, see https://github.com/intel/dffml/pull/694/files#r437573092

```suggestion
        if not level in current:
            current[level] = {}
        current = current[level]
```

dffml/util/data.py
```diff
@@ -109,13 +131,47 @@ def traverse_get(target, *args):
...
 
+def traverse_set(target, *args, value):
+    """
+    Examples
+    --------
+
+    >>> from dffml import traverse_set
+    >>>
+    >>> d = {"one": {"two": 3}}
+    >>> traverse_set(d,"one.two", value = "Three")
```
pdxjohnny, line 155: Space (style)

```suggestion
    >>> traverse_set(d, "one.two", value = "Three")
```

dffml/util/data.py
```diff
@@ -109,13 +131,47 @@ def traverse_get(target, *args):
...
+    >>> d["one"]["two"]
+    'Three'
```
pdxjohnny, line 157: Test for creating levels

```suggestion
    'Three'
    >>> traverse_set(d, "newdict.'other.value'", value = 42)
    >>> d["newdict"]
    {'other.value': 42}
```

PR #702, opened 2020-06-11
model: transformers: ner: Update NER model

0dust:
Rearranges and cleans NER model to support transformers 2.11.0


model/transformers/dffml_model_transformers/ner/utils.py
```diff
@@ -159,22 +171,10 @@ def convert_examples_to_features(
             segment_ids += [pad_token_segment_id] * padding_length
             label_ids += [pad_token_label_id] * padding_length
 
-        if not (len(input_ids) == max_seq_length):
```
pdxjohnny, line 162: Add these back in

PR #703, opened 2020-06-11
dffml: model: autosklearn: Add the autoclassifier model

sk-ip:


model/autosklearn/.coveragerc
```diff
@@ -0,0 +1,13 @@
+[run]
+source =
+    dffml_model_myslr
```
pdxjohnny, line 3:
```suggestion
    dffml_model_autosklearn
```

model/autosklearn/dffml_model_autosklearn/__init__.py
```diff
@@ -0,0 +1,15 @@
+"""
+**Ubuntu Installation**
+
+To provide a C++11 building environment and the lateste SWIG version on Ubuntu, run:
+
+.. code-block:: console
```
pdxjohnny, line 6:
```suggestion
.. code-block:: console

```

model/autosklearn/setup_common.py
```diff
@@ -0,0 +1,78 @@
...
+from pathlib import Path
+from setuptools import find_packages
+
+ORG = "dffml"
+NAME = "dffml-model-autosklearn"
+DESCRIPTION = "DFFML model dffml-model-autosklearn"
+AUTHOR_NAME = "Sudhanshu kumar"
+AUTHOR_EMAIL = "sudhanshukumar5459@gmail.com"
+# Install dffml if it is not installed in development mode
+INSTALL_REQUIRES = ["auto-sklearn"] + (
```
pdxjohnny, line 13:
```suggestion
INSTALL_REQUIRES = ["auto-sklearn>=0.7.0"] + (
```

model/autosklearn/dffml_model_autosklearn/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+    $ sudo apt-get install build-essential swig
+
+Install other PyPi dependencies with
```
pdxjohnny, line 9: You shouldn't need this section on the `requirements.txt`, those should be installed when the pypi package of `auto-sklearn` is installed

model/autosklearn/dffml_model_autosklearn/__init__.py
```diff
@@ -0,0 +1,15 @@
+"""
+**Ubuntu Installation**
```
pdxjohnny, line 2:
```suggestion
.. warning:

    Follow these instructions before running the above install
    command to ensure that ``auto-sklearn`` installs correctly

**Ubuntu Installation**
```

model/autosklearn/setup.py
```diff
@@ -0,0 +1,16 @@
...
+
+# Boilerplate to load commonalities
+spec = importlib.util.spec_from_file_location(
+    "setup_common", os.path.join(os.path.dirname(__file__), "setup_common.py")
+)
+common = importlib.util.module_from_spec(spec)
+spec.loader.exec_module(common)
+
+common.KWARGS["entry_points"] = {
+    "dffml.model": [f"misc = {common.IMPORT_NAME}.misc:MiscModel"]
```
pdxjohnny, line 13: This will of course need updating

sk-ip: yes. for now i have removed it. i will add to it once some models are implemented

model/autosklearn/dffml_model_autosklearn/__init__.py
sk-ip: yes, ideally that should happen,
but what happens is it would throw a lot of error on the screen and many wheels installation fails, pip again tries to build it from scratch then it gets installed. (installation does takes place but it throws a lot of errors and takes time)

installing it's dependencies first and then installing the auto-sklearn looks clean and works fast too.

So i thought maybe mention otherwise it works fine both ways.

pdxjohnny: Okay great. That's odd, good job finding this workaround that makes it clean!

model/autosklearn/dffml_model_autosklearn/config.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+    exclude_preprocessors: list = field(
+        "Specify the set of preprocessors not to use", default=None
+    )
+    resampling_strategy: str = field(
+        "How to to handle overfitting", default="holdout"
+    )
+    resampling_strategy_arguments: dict = field(
+        "Additional arguments for resampling_strategy",
+        default_factory={"train_size default": 0.67},
```
pdxjohnny, line 65:
```suggestion
        default_factory=lambda: {"train_size default": 0.67},
```

sk-ip: thanks

model/autosklearn/dffml_model_autosklearn/config.py
```diff
@@ -1,17 +1,23 @@
...
-from dffml import Feature, Features, Sources
 
 
 @config
 class AutoSklearnConfig:
     predict: Feature = field("Label or the value to be predicted")
     features: Features = field("Features on which to train the model")
+    directory: pathlib.Path = field(
+        "Directory where model should be saved",
+        default=pathlib.Path("~", ".cache", "dffml", "autosklearn"),
```
pdxjohnny, line 19: We have gone through and removed all the default directories for models. So let's not add a default here.

```suggestion
```

model/autosklearn/dffml_model_autosklearn/config.py
```diff
@@ -106,9 +112,15 @@ class AutoSklearnModelContext(ModelContext):
 
     def __init__(self, parent):
         super().__init__(parent)
-        self.pandas = importlib.import_module("pandas")
-        self.features = self._get_feature_names()
         self._model = None
+        self.pd = importlib.import_module("pandas")
```
pdxjohnny, line 116: It was mentioned in the last sync meeting that `importlib` messes with IDE's if you want to use it that's fine. But we don't have to use it.

sk-ip: Oh! ok

model/autosklearn/dffml_model_autosklearn/autoclassifier.py
```diff
@@ -0,0 +1,38 @@
...
+        super().__init__(parent)
+
+    @property
+    def model(self):
+        """
+        Generates or loads a model
+        """
+        if self._model is not None:
+            return self._model
+        self._model = autosklearn.classification.AutoSklearnClassifier(
```
pdxjohnny, line 22: https://github.com/intel/dffml/blob/b1906cb37ed4305ebcc0a1f3c0957daa1234c784/model/scikit/dffml_model_scikit/scikit_base.py#L79-L83

sk-ip: thanks for this :)

model/autosklearn/setup.py
```diff
@@ -0,0 +1,18 @@
...
+import importlib.util
+from setuptools import setup
+
+# Boilerplate to load commonalities
+spec = importlib.util.spec_from_file_location(
+    "setup_common", os.path.join(os.path.dirname(__file__), "setup_common.py")
+)
+common = importlib.util.module_from_spec(spec)
+spec.loader.exec_module(common)
+
```
pdxjohnny, line 11:
```suggestion

common.KWARGS["install_requires"] += ["auto-sklearn>=0.7.0"]
```

scripts/gitpod.Dockerfile
```diff
@@ -7,7 +7,7 @@ RUN echo 'unset PIP_USER' >> ~/.bashrc
 USER root
 
 RUN apt-get -q update && \
-  apt-get install -yq tmux libboost-all-dev && \
+  apt-get install -yq tmux libboost-all-dev build-essential swig && \
```
pdxjohnny, line 10: We need to add these to `.ci/deps.sh`

model/autosklearn/setup.py
sk-ip: done.

.ci/deps.sh
```diff
@@ -89,6 +89,11 @@ if [ "x${PLUGIN}" == "xmodel/tensorflow_hub" ]; then
   python -m pip install -U -e "./model/tensorflow"
 fi
 
+if [ "x${PLUGIN}" == "xmodel/autosklearn" ]; then
```
pdxjohnny, line 92: The pattern we follow here is to install the dependencies if we are running just the tests for the model, `if [[ "x${PLUGIN}" == "xmodel/autosklearn" ]]` or the main package `[[ "x${PLUGIN}" == "x." ]]` or the docs `[[ "x${PLUGIN}" == "xdocs" ]]`

This is because the main package will run `dffml service dev install`, which installs all plugins in development mode. And the docs will too, since we need to have it installed to check which models it provides so we can auto generate the entries on the models plugin page.

```suggestion
if [[ "x${PLUGIN}" == "xmodel/autosklearn" ]] || \
   [[ "x${PLUGIN}" == "x." ]] || \
   [[ "x${PLUGIN}" == "xdocs" ]]; then
```

sk-ip: Thanks

Issue #612, opened 2020-05-05
record: Improve string representation


pdxjohnny:
Make the `__str__` method output something more user friendly

pdxjohnny:
We should look at terminal width and create little ASCII tables for each record

Example ASCII table:

```
+=================================================================+
|   CVE Binary Tool Report Generated: 2020-04-29  10:04:06        |
+=================================================================+

+=================================================================+
|   MODULE NAME      |  VERSION  |    CVE NUMBER      | SEVERITY  |
+=================================================================+
| bzip2              | 1.0.2     | CVE-2005-0953      | LOW       |
+--------------------+-----------+--------------------+-----------+
| bzip2              | 1.0.2     | CVE-2005-1260      | MEDIUM    |
+--------------------+-----------+--------------------+-----------+
| bzip2              | 1.0.2     | CVE-2008-1372      | MEDIUM    |
+--------------------+-----------+--------------------+-----------+
```

pdxjohnny:
We probably want to have headers for fields going down the side, and truncate long fields that are going to be larger than width of terminal with `...`

pdxjohnny:
Within `======` top box we'd do the `record.key` and then going down the sides, we'd do feature data, then `++++` (or something that looks good) then prediction data.

Only ASCII, no UTF-8

pdxjohnny:
You will probably want to make another method that the `__str__` method calls, the new method will contain the pretty printing implementation, and be able to take keyword arguments, such as one to say which features not to print.

Issue #335, opened 2020-01-21
util: cli: cmd: Add ability to output in different formats and different amounts of data


pdxjohnny:
Right now output for prediction is dumping json of repo objects to stdout. We should improve this to not dump feature data if requested. We should also support output formats other than json.

sakshamarora1:
ping @pdxjohnny 
Also related issues: #610 #612

PR #708, opened 2020-06-12
record: Improve string representation

sakshamarora1:
Fixes: #612 
Fixes: #335 

```console
dffml predict all \
    -model tfdnnc \
    -model-batchsize 1000 \
    -model-hidden 30 50 25 \
    -model-clstype int \
    -model-predict label:int:1 \
    -model-classifications $(seq 0 9) \
    -model-features image:int:$((28 * 28)) \
    -sources images=csv \
    -source-filename image.csv  \
    -source-loadfiles image  \
    -pretty
    -log debug
```
Example output of 1 of the records:
```console
	Key:	four
                                                               Record Features
+----------------------------------------------------------------------------------------------------------------------------------------------+
|               image               |                 0,0,0,0,0,0,0,0,0,1,0,2,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,... (length:784)                 |
+----------------------------------------------------------------------------------------------------------------------------------------------+

                                                                  Prediction
+----------------------------------------------------------------------------------------------------------------------------------------------+
|                                                                    label                                                                     |
+----------------------------------------------------------------------------------------------------------------------------------------------+
|             Value:  3             |                                    Confidence:   0.13614192605018616                                     |
+----------------------------------------------------------------------------------------------------------------------------------------------+

```

For list records:
```console
dffml list records -sources f=csv -source-filename test.csv -pretty

	Key:	0
                                                               Record Features
+----------------------------------------------------------------------------------------------------------------------------------------------+
|             Expertise             |                                                    9                                                     |
+----------------------------------------------------------------------------------------------------------------------------------------------+
|               Salary              |                                                    50                                                    |
+----------------------------------------------------------------------------------------------------------------------------------------------+
|               Trust               |                                                   0.5                                                    |
+----------------------------------------------------------------------------------------------------------------------------------------------+
|               Years               |                                                    4                                                     |
+----------------------------------------------------------------------------------------------------------------------------------------------+
                                                                                                                   Prediction:    Undetermined

```

Signed-off-by: sakshamarora1 <sakshamarora1001@gmail.com>

### DIFF
--- a/dffml/record.py
+++ b/dffml/record.py
@@ -3,10 +3,12 @@
 """
 Information on the software to evaluate is stored in a Record instance.
 """
+import os
 from datetime import datetime
 from typing import Optional, List, Dict, Any
 
-from .util.data import merge
+from .util.data import merge, export
+from .util.display import create_row
 from .log import LOGGER
 
 LOGGER = LOGGER.getChild("record")
@@ -141,31 +143,35 @@ class Record(object):
         return str(self.dict())
 
     def __str__(self):
-        header = self.key
+        try:
+            width = int(3 * os.get_terminal_size().columns / 4)
+        except OSError:
+            width = 70
+        header = "\n\tKey:\t" + self.key
+        divider = "+" + "-" * (width) + "+"
+
         if len(self.extra.keys()):
-            header += " " + str(self.extra)
+            header += "\n\t" + str(self.extra)
 
         return "\n".join(
             [header]
+            + ["Record Features".center(width).rstrip()]
+            + [divider]
             + [
-                ("%-30s%s" % (feature, str(results)))
-                for feature, results in self.features().items()
+                create_row(feature, results, width)
+                for feature, results in map(
+                    lambda k: (k[0], export(k[1])), self.features().items()
+                )
             ]
-            + ["Predictions"]
             + (
-                [
-                    (
-                        "%-30s\n\tvalue:%s, confidence:%s"
-                        % (
-                            pred,
-                            str(conf_val["value"]),
-                            str(conf_val["confidence"]),
-                        )
-                    )
+                ["\n" + "Prediction".center(width).rstrip()]
+                + [divider]
+                + [
+                    create_row(pred, conf_val, width)
                     for pred, conf_val in self.data.prediction.items()
                 ]
                 if self.data.prediction
-                else ["Undetermined"]
+                else ["Prediction:    Undetermined".rjust(width)]
             )
         ).rstrip()
 
--- a/dffml/util/data.py
+++ b/dffml/util/data.py
@@ -215,6 +215,10 @@ def export_value(obj, key, value):
         obj[key] = value.export()
     elif hasattr(value, "_asdict"):
         obj[key] = value._asdict()
+    elif getattr(
+        type(value), "__module__", None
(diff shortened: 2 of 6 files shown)


sakshamarora1:
@pdxjohnny I added numpy in record.py for dealing with ndarrays of images later that is why half the tests are failing.
I think you told me something about not wanting numpy as a requirement at the top level? Sorry, I don't remember completely.

If I remove checking for ndarray and and change the array to tuple/list after using the png configloader then for preprocessing operations I will have to do np.array(<input_array>) each time at the beginning. What do you think?

dffml/record.py
```diff
@@ -3,6 +3,8 @@
 """
 Information on the software to evaluate is stored in a Record instance.
 """
+import os
+import numpy as np
```
pdxjohnny, line 7: Yes you were correct, we want to keep the top level package free of dependencies

```suggestion
```

dffml/record.py
```diff
@@ -141,31 +143,75 @@ def __repr__(self):
...
+                        + "|"
+                        + (
+                            ", ".join(
+                                str(val) for val in results[: int(width / 10)]
+                            )
+                            + f",... (length:{len(results)})"
+                        ).center(int(3 * width / 4))
+                        + "|"
+                    )
+                    if isinstance(results, (list, tuple, np.ndarray,))
```
pdxjohnny, line 175:
```suggestion
                    if isinstance(results, (list, tuple,))
```

dffml/record.py
```diff
@@ -141,31 +143,75 @@ def __repr__(self):
...
+                        + "|"
+                        + str(results).center(int(3 * width / 4))
+                        + "|"
+                    )
+                    if isinstance(results, (str, int, float, bool,))
+                    else ""
+                )
+                + "\n"
+                + divider
                 for feature, results in self.features().items()
```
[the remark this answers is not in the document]
pdxjohnny: If this suggestion results in an error then you'll want to go modify `export_value` in `dffml.util.data`

dffml/record.py
```diff
@@ -141,31 +143,75 @@ def __repr__(self):
...
-            header += " " + str(self.extra)
+            header += "\n\t" + str(self.extra)
 
         return "\n".join(
             [header]
+            + [feature_header]
+            + [divider]
             + [
-                ("%-30s%s" % (feature, str(results)))
+                (
```
pdxjohnny, line 162: This "building a table from a dict" should be it's own function. Maybe it should go somewhere like `dffml.util.display`? `display` may not be the best word choice, if you think of something else go for it.

dffml/util/data.py
```diff
@@ -215,6 +215,8 @@ def export_value(obj, key, value):
         obj[key] = value.export()
     elif hasattr(value, "_asdict"):
         obj[key] = value._asdict()
+    elif getattr(type(value), "__module__", None) == "numpy":
```
pdxjohnny, line 218: We should also check that the `flatten` method exists and can be called to be safe

```suggestion
    elif getattr(type(value), "__module__", None) == "numpy" and inspect.ismethod(getattr(value, "flatten", None)):
```

sakshamarora1: Alright!

PR #719, opened 2020-06-19
WIP: model: transformers: qa: Add Question Answering model

0dust:


model/transformers/examples/qa/train_data.sh
```diff
@@ -0,0 +1,66 @@
+cat >> train.json << EOF
+{
+    "data":[
+       {
+          "title":"University_of_Notre_Dame",
+          "paragraphs":[
+             {
+                "context":"Architecturally, the school has a Catholic character. Atop the Main Building's gold dome is a golden statue of the Virgin Mary. Immediately in front of the Main Building and facing it, is a copper statue of Christ with arms upraised with the legend \"Venite Ad Me Omnes\". Next to the Main Building is the Basilica of the Sacred Heart. Immediately behind the basilica is the Grotto, a Marian place of prayer and reflection. It is a replica of the grotto at Lourdes, France where the Virgin Mary reputedly appeared to Saint Bernadette Soubirous in 1858. At the end of the main drive (and in a direct line that connects through 3 statues and the Gold Dome), is a simple, modern stone statue of Mary.",
```
pdxjohnny, line 8: It looks like this is from a book. We can't use something that might be copyrighted (and I'd assume is in this case, unless you find out otherwise)

model/transformers/examples/qa/accuracy.sh
```diff
@@ -0,0 +1,11 @@
+dffml accuracy \
+  -model qa_model \
+  -model-model_type bert \
+  -model-save_steps 3 \
+  -model-model_name_or_path bert-base-cased \
+  -model-output_dir qamodel/checkpoints \
+  -model-cache_dir qamodel/cache \
+  -sources s=op \
+  -source-opimp model.transformers.dffml_model_transformers.qa.utils:parser \
```
pdxjohnny, line 9: We should be able to load it with this syntax, the syntax you've used assumes the user is running from the root of the DFFML source tree.

```suggestion
  -source-opimp dffml_model_transformers.qa.utils:parser \
```

model/transformers/dffml_model_transformers/qa/qa_model.py
```diff
@@ -0,0 +1,1006 @@
...
+            )
+            record.predicted("Answer", prediction, "Nan")
+            yield record
+
+
+@entrypoint("qa_model")
+class QAModel(Model):
+    """
+    Implemented using `HuggingFace Transformers <https://huggingface.co/transformers/index.html>`_ Pytorch based Models.
+    Description about pretrianed models can be found `here <https://huggingface.co/transformers/pretrained_models.html>`_
```
pdxjohnny, line 939: We should really `literalinclude` the `parser.py` code in here and talk about how we're using it to convert the format of the dataset. It might go best at the end.

model/transformers/dffml_model_transformers/qa/qa_model.py
```diff
@@ -0,0 +1,1006 @@
...
+        # Check if continuing training from a checkpoint
+        if os.path.exists(self.parent.config.model_name_or_path):
+            try:
+                # set global_step to gobal_step of last saved checkpoint from model path
+                checkpoint_suffix = self.parent.config.model_name_or_path.split(
+                    "-"
+                )[
+                    -1
+                ].split(
+                    "/"
```
pdxjohnny, line 387: This looks like it assumes UNIX paths, can we use `pathlib.Path` here instead to make sure this will still work on Windows?

model/transformers/examples/qa/train_data.sh
0dust: Oh! This is from SQuAD dataset distributed under CC-BY-SA-4.0 license. Is that problematic? If so, I can make up something to write here.

pdxjohnny: That's perfect! We just need to document that somewhere

model/transformers/dffml_model_transformers/qa/qa_model.py
```diff
@@ -1000,6 +999,25 @@ class QAModel(Model):
...
+    The location of the function is passed using: 
+
+    .. code-block:: console
+
+            -source-opimp dffml_model_transformers.qa.utils:parser
+    And the arguments to `parser` are passed by:
+
+    .. code-block:: console
+
+            -source-args train.json True
```
pdxjohnny, line 1013:
```suggestion
            -source-args train.json True

```

model/transformers/dffml_model_transformers/qa/qa_model.py
```diff
@@ -1000,6 +999,25 @@ class QAModel(Model):
...
+    read and parse data from json file before feeding it to the model. The function used by opsource to parse json data
+    is:
+
+    .. literalinclude:: /../model/transformers/dffml_model_transformers/qa/utils.py
+
+    The location of the function is passed using: 
+
+    .. code-block:: console
+
+            -source-opimp dffml_model_transformers.qa.utils:parser
```
pdxjohnny, line 1008:
```suggestion
            -source-opimp dffml_model_transformers.qa.utils:parser

```

PR #762, opened 2020-07-04
examples: ffmpeg: Remove files from when ffmpeg was a package

aghinsa:


aghinsa:
@pdxjohnny Sorry!! I remember doing all this, my guess is I messed up when resetting and force pushed a previous commit. :sweat_smile:

Issue #732, opened 2020-06-23
accuracy: Create new plugin type of scoring models


pdxjohnny:
We'll do a review at the end of each phase. For each phase, we'll open a PR against the `accuracy_scorer` branch (rather than `master`)

### Phase 1 - Initial Implementation

- [x] Create a new plugin type, Accuracy
- [x] Each entry will have a
  - [x] CONFIG
  - [x] Score coroutine method (`async def`)
    - [x] Inputs
      - [x] First is the same as the output of the predict method (an asynciterator of Record objects)
      - [x] Second is list of Feature instances that we did prediction on
    - [x] The method will look at the record.features() for the features we predicted, and the record.predictions() and return it's assessment of the accuracy.
- [x] Change accuracy method of model to no longer be an `abstractmethod`, we'll make it so that it takes a second parameter, the accuracy_scorer (of type `AccuracyContext`). The `accuracy` method within `ModelContext` will
  - [x] check if `self.config.predict` exists using `hasattr`, if not raise `NotImplementedError`
  - [x] return a call to `accuracy_scorer.score(self.predict(sources), [self.config.predict])`
- [x] We will need to modify the high level accuracy function to take accuracy_scorer as the second argument (after `model`)
- [x] We will need to implement at least one of these from scratch within the main package (dffml) so that we can use it for testing and examples and the like

### Phase 2 - Documentation

- [x] Add and example with tests for the MSE accuracy scoring plugin
- [x] Modify `dffml/model/slr.py` to remove the accuracy method
- [x] Modify `scripts/docs/care` to add a `dffml.accuracy dffml` line

### Phase 3 - Update Model Plugins

  - [x] Remove `accuracy` method from the following models
    - [x] dffml-model-scratch
    - [x] dffml-model-scikit
    - [x] dffml-model-tensorflow
    - [x] dffml-model-vowpalWabbit
    - [x] dffml-model-autosklearn
    - [x] dffml-model-daal4py
  - [x] For NLP models modify accuracy method signature to accept a scorer but log.warning that it won't be used
    - [x] dffml-model-tensorflow-hub
    - [x] dffml-model-transformers

### Phase 3.33 - Update Model Plugin - Merge New Models from master branch

- [x] Create PR that merges new models from master into accuracy scorer branch

### Phase 3.66 - Update Model Plugin - Again

  - [x] Remove `accuracy` method from the following models
    - [x] dffml-model-pytorch
[...]

PR #779, opened 2020-07-12
dffml: accuracy: Accuracy type plugin

sk-ip:
addresses #732

Signed-off-by:  Sudhanshu <Sudhanshukumar5459@gmail.com>


dffml/model/model.py
```diff
@@ -50,13 +51,20 @@ async def train(self, sources: Sources):
...
+    async def accuracy(
+        self, sources: Sources, accuracy_scorer: AccuracyContext
+    ) -> Accuracy:
         """
         Evaluates the accuracy of our model after training using the input records
         as test data.
         """
-        raise NotImplementedError()
+        if not hasattr(self.config, "predict") or not hasattr(
+            self.config, "features"
```
pdxjohnny, line 62: I don't think we need the `or features` here. Since we don't use it in this function

sk-ip: ok.

pdxjohnny:
Great work!!

PR #802, opened 2020-07-23
service:http: Add support for immediate response

### LINKED ISSUE #657: service: http: HTTPChannelConfig: Allow for immdiate responses
We need to have something like this in the channel config

```yaml
immidiate_repsonse:
  status: 200
  body:
    error: null
  headers:
    Content-Type: application/json
```

Make responds appropriately depending on content type (so JSON if
`application/json`, etc, default to `text/plain`)
If set, respond with the body, and status code given. And use
`asyncio.create_task` to kick off the dataflow (which will run in background,
since we used `create_task`).

We needs tests for this too of course

aghinsa:
fixes #657


service/http/docs/dataflow.rst
```diff
@@ -72,3 +72,23 @@ HttpChannelConfig
...
+        - headers
+        eg:
+
+        .. code-block:: python
+
+            {
+                "status": 200,
+                "content_type": "application/json",
+                "data": {"text": "ok"},
+            },
```
pdxjohnny, line 92:
```suggestion
            }
```

service/http/docs/dataflow.rst
```diff
@@ -72,3 +72,23 @@ HttpChannelConfig
...
+
```
pdxjohnny, line 93:
```suggestion
```

service/http/dffml_service_http/routes.py
```diff
@@ -342,6 +359,30 @@ async def multicomm_dataflow(self, config, request):
                         status=HTTPStatus.NOT_FOUND,
                     )
 
+    async def multicomm_dataflow(self, config, request):
+        if config.immediate_response:
+            asyncio.create_task(self._multicomm_dataflow(config, request))
+
+            ir = config.immediate_response
+            content_type = ir["content_type"]
```
pdxjohnny, line 367: We should auto determine the content type here if it's not given. For example

- If it doesn't exist and
  - data exists
    - application/json
  - text exists
    - text/plain
  - body exists
    - application/octet-stream

Issue #783, opened 2020-07-16
skel: Replace usage of pkg_resources with importlib.resources


pdxjohnny:
Just came across this: https://docs.python.org/3/library/importlib.html#importlib.resources.path

We should get rid of some usages of `pkg_resources` within our codebase. This would make it so we could be `zip_safe`!

Nishikoh:
Hi, I'd like to tackle that.

pdxjohnny:
@Nishikoh Awesome! Just let me know, or anyone on Gitter if you need more info. This should be a case of finding all the `pkg_resources.resource_filename` and changing them

```console
$ git grep pkg_resources | grep resource_
dffml/util/skel.py:    pkg_resources.resource_filename("dffml", "skel")
service/http/dffml_service_http/routes.py:    pkg_resources.resource_filename("dffml_service_http", "api.js")
```

PR #807, opened 2020-07-25
[WIP] skel: Replace pkg_resources.resource_filename

Nishikoh:
Part of issue https://github.com/intel/dffml/issues/783

### DIFF
--- a/dffml/util/skel.py
+++ b/dffml/util/skel.py
@@ -1,4 +1,4 @@
-import pkg_resources
+import importlib.resources
 from pathlib import Path
 import distutils.dir_util
 from typing import NamedTuple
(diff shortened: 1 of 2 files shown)


pdxjohnny:
@Nishikoh I was just about to to say that this looks tricky. I think I'd missed the part about `.path` not working on directories. I think you've got the `API_JS_BYTES` part correct, but we might be able to turn that into a one liner using [importlib.resources.read_binary](https://docs.python.org/3/library/importlib.html#importlib.resources.read_binary)

As for `skel`, I'm sadly not optimistic on that one. There seems to be no good way to iterate over resources within a subdirectory of a package (which is what we're looking for here).

Nishikoh:
@pdxjohnny  Do you think [this](https://docs.python.org/3/library/importlib.html#importlib.abc.ResourceReader.contents) will make it better?

pdxjohnny:
@Nishikoh The issue that you'll face with that is that

> The iterable does not recurse into subdirectories.

From: https://docs.python.org/3/library/importlib.html#importlib.resources.contents which is a wrapper around the calling of that method so far as I can tell

Also: https://github.com/python/cpython/blob/038827d0f59076f52e9797018ead12b1295cddc2/Lib/importlib/resources.py#L241-L259

Where it does `os.listdir`...

PR #808, opened 2020-07-25
dffml: model: Update the accuracy method from the models

sk-ip:
Addresses #732 

### DIFF
--- a/model/autosklearn/dffml_model_autosklearn/config.py
+++ b/model/autosklearn/dffml_model_autosklearn/config.py
@@ -167,21 +167,6 @@ class AutoSklearnModelContext(ModelContext):
             record.predicted(target, predict, max(prob))
             yield record
 
-    async def accuracy(self, sources: Sources) -> Accuracy:
-        if not self.model:
-            raise ModelNotTrained("Train the model before assessing accuracy")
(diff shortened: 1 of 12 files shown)


model/transformers/dffml_model_transformers/qa/qa_accuracy.py
```diff
@@ -0,0 +1,71 @@
...
+class QAAccuracyConfig:
+    pass
+
+
+class QAAccuracyContext(AccuracyContext):
+    async def score(
+        self, records: AsyncIterator[Record], features: List[Feature],
+    ):
+        if not os.path.isfile(
+            os.path.join(self.parent.config.output_dir, "pytorch_model.bin")
```
pdxjohnny, line 28: I'm realizing we don't have access to any of the model parameters. Sorry, we shouldn't have taken out these methods. I'm now remembering that's why we modified the accuracy method in the base model class. So that we could override it if necessary. We should revert these last six commits, and modify the signature of the accuracy method to match the new version that takes the scorer. However we should use `logger.warning` to mention that we aren't going to use the accuracy scorer provided

pdxjohnny:
Nice!

PR #829, opened 2020-08-04
model: Update the spacy, pytorch, xgboost models

sk-ip:
Addresses #732


sk-ip:
Oh! i didn't notice it, thanks for noticing it :) made the changes.

PR #832, opened 2020-08-18
WIP: dffml: tests: Update the tests

sk-ip:
Addresses #732


dffml/cli/ml.py
```diff
@@ -10,11 +10,15 @@
...
 from ..base import config, field
+from ..accuracy import AccuracyScorer, MeanSquaredErrorAccuracy
 
 
 @config
 class MLCMDConfig(SourcesCMDConfig, ModelCMDConfig):
-    pass
+    accuracy_scorer: AccuracyScorer = field(
+        "Method to use to score accuracy",
+        default_factory=lambda: MeanSquaredErrorAccuracy(),
```
pdxjohnny, line 20: I'm realizing that we should probably not have a default here, because we are not providing a default for the `high_level` functions. This would be an inconsistency.

```suggestion
```

model/autosklearn/examples/autoclassifier/accuracy.sh
```diff
@@ -9,4 +9,5 @@ dffml accuracy \
     SepalWidth:float:1 \
     PetalLength:float:1 \
     PetalWidth:float:1 \
+  -accuracy-scorer=mse \
```
pdxjohnny, line 12: We usually don't use `=`, so let's not use it for this.

```suggestion
  -accuracy-scorer mse \
```

Side note, you can look for commands that will need to be changed to add the `-accuracy-scorer` using a command like:

```console
$ git grep -A 25 -E 'dffml train|dffml accuracy|dffml predict' | python -c 'import sys; print("--".join([i for i in sys.stdin.read().split("--") if not "accuracy-scorer" in i]).strip())'
```

However, that will only show you `.sh` commands, you'll want maybe something like

```console
$ git grep -A 25 -E 'train|accuracy|predict'
```

To look at it all, but that's going to give you a ton of results to filter though

model/daal4py/tests/test_lr_integration.py
```diff
@@ -44,6 +44,8 @@ async def test_run(self):
         await CLI.cli(
             "accuracy",
             *model_args,
+            "-accuracy_scorer",
```
pdxjohnny, line 47:
```suggestion
            "-accuracy-scorer",
```

dffml/cli/ml.py
```diff
@@ -10,11 +10,15 @@
...
+    )
```
pdxjohnny, line 21: A few things I realized

- Since this will only be used for the `accuracy` command, we should make a separate `Config` for it.
- Since we won't provide a default for the function in `high_level` we probably shouldn't define it here, that would create an inconsistency
- We should probably shorten the name of the flag to `-scorer`, since that avoid the user having to type out `-accuracy` twice.

```suggestion
    pass

@config
class AccuracyCMDConfig(MLCMDConfig):
    scorer: AccuracyScorer = field(
        "Method to use to score accuracy",
    )
```

model/daal4py/examples/lr/lr.py
```diff
@@ -12,7 +13,10 @@
 train(model, "train.csv")
 
 # Assess accuracy (alternate way of specifying data source)
-print("Accuracy:", accuracy(model, CSVSource(filename="test.csv")))
+mse_accuracy = MeanSquaredErrorAccuracy()
+print(
+    "Accuracy:", accuracy(model, mse_accuracy, CSVSource(filename="test.csv"))
+)
```
pdxjohnny, line 19: Since we have the `model` argument named `model`, let's name the accuracy scorer `scorer`, for consistency.

```suggestion
scorer = MeanSquaredErrorAccuracy()
print(
    "Accuracy:", accuracy(model, scorer, CSVSource(filename="test.csv"))
)
```

model/xgboost/tests/test_model.py
```diff
@@ -73,10 +74,11 @@ async def test_00_train(self):
         await train(self.model, self.trainingsource)
 
     async def test_01_accuracy(self):
+        mse_accuracy = MeanSquaredErrorAccuracy()
         # Use the test data to assess the model's accuracy
-        res = await accuracy(self.model, self.testsource)
+        res = await accuracy(self.model, mse_accuracy, self.testsource)
         # Ensure the accuracy is above 80%
-        self.assertTrue(0.8 <= res)
+        self.assertTrue(res)
```
pdxjohnny, line 81: I think this might have been just failing periodically, is it failing all the time for you with the accuracy check?

dffml/cli/ml.py
```diff
@@ -46,16 +56,18 @@ class Accuracy(MLCMD):
     CONFIG = MLCMDConfig
```
sakshamarora1, line 56:
```suggestion
    CONFIG = AccuracyCMDConfig
```

sakshamarora1, line 56:
```suggestion
    CONFIG = AccuracyCMDConfig
```

model/autosklearn/dffml_model_autosklearn/config.py
```diff
@@ -122,9 +122,16 @@ def _get_feature_names(self):
...
+        try:
+            async for record in sources.with_features(
+                self.features + [self.parent.config.predict.name]
+            ):
+                ret_record.append(record)
+        except Exeption as e:
+            async for record in sources.with_features(self.features):
+                ret_record.append(record)
+        finally:
+            return ret_record
```
pdxjohnny, line 134:
```suggestion
        async for record in sources.with_features(
            self.features + [self.parent.config.predict.name] if with_predict else []
        ):
            ret_record.append(record)
        return ret_record
```

model/autosklearn/dffml_model_autosklearn/config.py
```diff
@@ -122,9 +122,16 @@ def _get_feature_names(self):
 
     async def get_test_records(self, sources: SourcesContext):
```
pdxjohnny, line 123:
```suggestion
    async def get_test_records(self, sources: SourcesContext, with_predict: bool):
```

dffml/accuracy/clfacc.py
```diff
@@ -0,0 +1,46 @@
+from typing import AsyncIterator, List
```
pdxjohnny, line 1: Looks great!

model/tensorflow/examples/tfdnnc/accuracy.sh
```diff
@@ -4,6 +4,7 @@ dffml accuracy \
   -model-directory tempdir \
   -model-classifications 0 1 2 \
   -model-clstype int \
+  -scorer clfacc \
```
pdxjohnny, line 7: - [ ] It might be good to just name this `clf`, since the `acc` is a bit redundant, and isn't present in `mse`

setup.py
```diff
@@ -134,7 +134,8 @@
         "dffml.secret": ["ini = dffml.secret.ini:INISecret"],
         # Accuracy
         "dffml.accuracy": [
-            "mse = dffml.accuracy.mse:MeanSquaredErrorAccuracy"
+            "mse = dffml.accuracy.mse:MeanSquaredErrorAccuracy",
+            "clfacc = dffml.accuracy.clfacc:ClassificationAccuracy",
```
pdxjohnny, line 138: `clfacc` -> `clf`, we'll need to change the filename too

```suggestion
            "clf = dffml.accuracy.clf:ClassificationAccuracy",
```

pdxjohnny:
Thank you so much for how you're doing these commits, it makes it so much easier to review!! Nice job :)

model/tensorflow_hub/dffml_model_tensorflow_hub/text_classifier_accuracy.py
```diff
@@ -24,19 +25,17 @@ class TextClassifierAccuracyContext(AccuracyContext):
...
         if not os.path.isfile(
-            os.path.join(self.model_dir_path, "saved_model.pb")
+            os.path.join(modelcontext.model_dir_path, "saved_model.pb")
         ):
             raise ModelNotTrained("Train model before assessing for accuracy.")
-        x, y = await self.train_data_generator(records)
-        accuracy_score = self._model.evaluate(x, y)
+        x, y = await modelcontext.train_data_generator(sources)
+        accuracy_score = modelcontext._model.evaluate(x, y)
         return accuracy_score[1]
```
pdxjohnny, line 35: The implementation of this makes me think that we might want to remove the accuracy method from the models entirely. We should talk more about this in the meeting, and we may not necessarily need to do it within phase 4, it could wait.

I say this because we seem to have a few cases now where the accuracy scorer might needs more information from / about the model. But the only case we have where the model needs more information about the accuracy scorer is when we're saying that a certain accuracy scorer won't work or won't be used.

pdxjohnny:
Fantastic work!!

PR #843, opened 2020-09-02
df: input flow: Add get_alternate_definitions helper function

pdxjohnny:
We had a lack of clarity around what was going on with alternate
definitions in commit 2ed6e172669f9f9c15c96409ea5ed5673c5b6c8b. There
were several if statements where we are looping through the origins
and if the origin is a dict, we check if the value is a tuple or a list.

If it is not, the key is the operation instance name and the value is
the output name for that operation instance.

If it is, the key is the origin and the value is the list or tuple of
definition names which are acceptable alternate definitions when the
input is coming from that origin.

This patch adds a helper method to make that more clear.


pdxjohnny:
[transformers.log](https://github.com/intel/dffml/files/5163619/transformers.log)

PR #862, opened 2020-10-01
Fixes #834

### LINKED ISSUE #834: source: file: close does not use WRITEMODE
https://github.com/intel/dffml/blob/78f7ba8de290214919dc45de50800030bbe60f24/dffml/source/file.py#L108

does not use `self.WRITEMODE` (unlike `open`)

change the `"w+"` to be `self.WRITEMODE` and then fix the associated tests

aadarshsingh191198:


pdxjohnny:
Thanks @aadarshsingh191198! Can you run `black` on this?

```console
$ pip install black==19.10b0
$ black .
```

aadarshsingh191198:
Sure @pdxjohnny. Will do!!

PR #864, opened 2020-10-07
Fixes #651

### LINKED ISSUE #651: model: scikit: Data type of feature used for predict not respected
Predicted values are comming out as floats from regression models. We should
typecast to the feature's dtype.

aadarshsingh191198:


pdxjohnny:
This doesn't seem to be working.

You're interested in making sure the `model/scikit` tests work:

```console
$ (cd model/scikit && python setup.py test)
```

PR #878, opened 2020-11-03
WIP: setup: notify incompatible Python at installation

ichisadashioko:
These changes checks Python version at installation and throws exception if current Python version is less than 3.7. This is because we are using Python 3.7 features which are not available at older versions (e.g. `AsyncExitStack` in `dffml/util/asynchelper.py`).

example in Colab

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,11 +1,20 @@
 # SPDX-License-Identifier: MIT
 # Copyright (c) 2019 Intel Corporation
 import os
+import sys
 import ast
 import pathlib
 from io import open
 import importlib.util
 from setuptools import find_packages, setup
+
+
+class InstallException(Exception):
+    pass
+
+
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -1,12 +1,21 @@
...
 import importlib.util
 from setuptools import find_packages, setup
 
+
+class InstallException(Exception):
+    pass
+
+
+if sys.version_info.major != 3 and sys.version_info.minor < 7:
+    raise InstallException("dffml is incompatible with Python version <= 3.7!")
```
pdxjohnny, line 17:
```suggestion
    raise InstallException("dffml is incompatible with Python version < 3.7!")
```

pdxjohnny (changes requested):
With just this edit I think that we are good to go on this PR. Thanks!

PR #900, opened 2020-11-20
plugins: Added requirements.txt for each plugin

### LINKED ISSUE #890: plugins: Move install_requires into requirements.txt
We need to go through all the `setup.py` and `setup_common.py` for each plugin and take anything in the `install_requires` and move it to a `requirements.txt` file within the same directory. Then we need to modify `setup_common.py` (or `setup.py` if there is no `setup_common.py`) similar to how was done in 4867b4c1d6b1a51ef89e3926d0561a5502ad2850

```python
REQUIREMENTS_TXT_PATH = Path(SELF_PATH, "requirements.txt")
if REQUIREMENTS_TXT_PATH.is_file():
    INSTALL_REQUIRES += list(
        map(lambda i: i.strip(), REQUIREMENTS_TXT_PATH.read_text().split("\n"))
    )
```

SGeetansh:
Fixes #890


pdxjohnny:
Nice! Thank you!

PR #911, opened 2020-11-29
Formatted the HTML, CSS and JS files in documentation

nitesh585:
I have observed that code is not in the right format and sometimes it looks like a huge chunk So, I used HTMLpretiffy to format the HTML, CSS, and js files.


pdxjohnny:
HTML docs are generated via [sphinx](https://www.sphinx-doc.org/en/3.x/contents.html) so they'll get re-built to their unformated version every time. This isn't something that's worth spending time on unfortunatly, and if it was, the right place to do it would be in Sphinx: https://github.com/sphinx-doc/sphinx/blob/3.x/sphinx/builders/html/__init__.py#L170

nitesh585:
Actually I am beginner to open source.

Thank you so much for your guidance. Now, I will try to solve some other genuine issues.

pdxjohnny:
Sounds good! Try filtering by [good first issue](https://github.com/intel/dffml/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22)

Issue #910, opened 2020-11-24
model: pytorch: Custom exception when loss function is misspelled


pdxjohnny:
Need helpful error messages when loss function is misspelled, try https://intel.github.io/dffml/master/tutorials/neuralnetwork.html with a misspelling of the loss function to see the current error.

PR #915, opened 2020-12-04
model: pytorch: Custom exception when loss function is misspelled

nitesh585:
Solved issue #910.

**Exception message:**
Before change - 'Nonetype' object has no attribute 'function'
After change- Loss function is not valid!

### DIFF
--- a/model/pytorch/dffml_model_pytorch/pytorch_base.py
+++ b/model/pytorch/dffml_model_pytorch/pytorch_base.py
@@ -63,6 +63,14 @@ class PyTorchModelConfig:
                 map(self.clstype, self.classifications)
             )
 
+class LossFunctionNotFoundError(Exception):
+    """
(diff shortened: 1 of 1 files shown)


model/pytorch/dffml_model_pytorch/pytorch_base.py
```diff
@@ -63,6 +63,14 @@ def __post_init__(self):
                 map(self.clstype, self.classifications)
             )
 
+class LossFunctionNotFoundError(Exception):
+    """
+    Exception raised when loss function is misspelled.
+    """
+    def __init__(self, message="Loss function is not valid!"):
```
pdxjohnny, line 70: Not sure if we need this `__init__`. It would probably be sufficient to just pass this error message every time. We should probably say what the spelling they gave was, if possible.

Issue #648, opened 2020-05-19
util: packaging: is_develop should check if package name with _ or - has .egg-link


pdxjohnny:
https://github.com/intel/dffml/blob/23064bcb5de738b98007ffc3225516c24eb27d84/dffml/util/packaging.py#L15-L18

aadarshsingh191198:
@pdxjohnny I want to work on this but am not sure as to what exactly is wanted. Could you please guide me?

pdxjohnny:
@aadarshsingh191198 Python will sometimes create `.egg-link` files for packages that contain `-` in the name where the `.egg-link` file has the package name but with `-` replaced with `_`.

We just need to put the above block into a loop were the first iteration of the loop just does the check as it does it now, and the second iteration checks for the presence of the `.egg-link` file but with the package name where `-` has been replaced with `_`. This will ensure that if Python installed it with `_` in the name, we still detect it successfully.

aadarshsingh191198:
Thanks ... Now I understood the problem. Will start working on it asap!!

aitikgupta:
I see the check still hasn't been added yet. Is it okay if I take this up?
/cc @aadarshsingh191198 @pdxjohnny

aadarshsingh191198:
Ya sure you can. I got busy in some work so will not be able to take this right now.

Issue #619, opened 2020-05-08
util: testing: Move core docstrings test logic into main package


pdxjohnny:
We should refactor `tests/test_docstrings.py` into `dffml/util/testing/docstrings.py` and make it so that the plugins all have a file under their tests which imports and uses the docstring test code so that the plugins docstrings are always tested.

aitikgupta:
Is it okay if I take this up?
Some pointers would do me good, I looked around `tests/test_docstrings.py`, the flow is somewhat clear, but the proposed util: "so that plugins have a file under their tests" isn't very clear to me.
Any guide would help :)

pdxjohnny:
@aitikgupta I've updated the issue body, sorry for the slow response! Let me know if you have further questions.

You may have moved on to other things, so for anyone else reading this, the issue is up for grabs, subject to the contributing guidelines: https://intel.github.io/dffml/master/contributing/dev_community.html#communication-style

programmer290399:
I'm still working on this.

PR #932, opened 2020-12-17
Check for alternate package name (-, _) having .egg-link

aitikgupta:
This PR addresses #648:
Adds a check if a package name has `_` instead of `-` with a `.egg-link`.


pdxjohnny:
Sorry I've been on vacation. Looks great! Thank you!

aitikgupta:
That's okay!
While I have you - could you take a look at https://github.com/intel/dffml/issues/619#issuecomment-748223882? 😄

PR #980, opened 2021-02-06
Cast predict to input dtype

### LINKED ISSUE #651: model: scikit: Data type of feature used for predict not respected

### LINKED ISSUE #978: df: base: op: Add keyword argument to toggle interpretation of None
Add a boolean keyword argument to [`@op`](https://intel.github.io/dffml/master/api/df/base.html?highlight=op#dffml.df.base.op) to toggle interpretation of the wrapped function's return value.

`valid_return_none` might be a good choice. If `valid_return_none` is true. Then we should build the result dict

References:
- 51d5a979dcb3e54ab970c5fc84b56d62cd908bcf

sidhu1012:
Fixes #651

Predicted values are type casted to input type if input isn't `str`.

### DIFF
--- a/model/scikit/dffml_model_scikit/scikit_base.py
+++ b/model/scikit/dffml_model_scikit/scikit_base.py
@@ -148,7 +148,10 @@ class ScikitContext(ModelContext):
             )
             target = self.parent.config.predict.name
             record.predicted(
-                target, self.clf.predict(predict)[0], self.confidence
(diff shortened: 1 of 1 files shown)


sidhu1012:
Are some tests going to fail certainly?

PR #985, opened 2021-02-11
Add `valid_return_none` argument to @op

### LINKED ISSUE #978: df: base: op: Add keyword argument to toggle interpretation of None

sidhu1012:
Fixes #978

### DIFF
--- a/dffml/df/base.py
+++ b/dffml/df/base.py
@@ -252,7 +252,7 @@ def create_definition(name, param_annotation, default=NO_DEFAULT):
     )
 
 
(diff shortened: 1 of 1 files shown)


pdxjohnny:
Looks like we need to run `black`. See https://intel.github.io/dffml/master/contributing/style.html

PR #1001, opened 2021-02-18
util: net: cached_download_unpack_archive: Remove directory on failur…

### LINKED ISSUE #867: util: net: cached_download_unpack_archive: Remove directory on failure to extract
If we fail to extract the archive. We should remove the directory we were going to extract to.

https://github.com/intel/dffml/blob/3a73d521da2d73b0dd9fbe926af41b737aba835c/dffml/util/net.py#L141-L219

nitesh585:
fix #867

### DIFF
--- a/dffml/util/net.py
+++ b/dffml/util/net.py
@@ -1,3 +1,5 @@
+import stat
+import os
 import shutil
 import hashlib
 import pathlib
@@ -35,6 +37,19 @@ class ProtocolNotAllowedError(Exception):
 
     def __str__(self):
         return f"Protocol of URL {self.url!r} is not in allowlist: {self.allowlist!r}"
+
+
+class DirectoryNotExtractedError(Exception):
+    """
+    Raised when extraction of directory failed!
+    """
+
+    def __init__(self, directory_path):
+        super().__init__()
+        self.directory_path = directory_path
+
+    def __str__(self):
+        return f"Failed to extract - {self.directory_path!r}"
 
 
 # Default list of URL protocols allowed
@@ -195,12 +210,27 @@ def cached_download_unpack_archive(
     >>> asyncio.run(files_in_dffml_commit_c4469a())
     124
     """
+
+    def on_error(func, path, exc_info):
+        """
+        Error handler for shutil.rmtree
+        """
+        if not os.access(path, os.W_OK):
+            os.chmod(path, stat.S_IWUSR)
+            func(path)
+        else:
+            pass
+
     directory_path = pathlib.Path(directory_path)
 
     async def extractor(download_path):
         download_path = download_path.absolute()
         with chdir(directory_path):
-            shutil.unpack_archive(str(download_path), ".")
+            try:
+                shutil.unpack_archive(str(download_path), ".")
+            except Exception as error:
+                shutil.rmtree(directory_path, onerror=on_error)
+                raise DirectoryNotExtractedError(directory_path) from error
 
     extract = cached_download(
         url, file_path, expected_hash, protocol_allowlist=protocol_allowlist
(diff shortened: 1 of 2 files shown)


dffml/util/net.py
```diff
@@ -37,6 +40,19 @@ def __str__(self):
...
+    """
+    Raised when extraction of directory failed!
+    """
+
+    def __init__(self, directory_path):
+        super().__init__()
+        self.directory_path = directory_path
+
+    def __str__(self):
+        return f"Failed to extract - {self.directory_path}."
```
pdxjohnny, line 53: - Exceptions usually don't end in periods
- With file paths it can be helpful to print with `repr()`, in case there are spaces or other odd characters the user may be having trouble noticing.

```suggestion
        return f"Failed to extract - {self.directory_path!r}"
```

dffml/util/net.py
```diff
@@ -195,12 +211,26 @@ def cached_download_unpack_archive(
...
     directory_path = pathlib.Path(directory_path)
 
     async def extractor(download_path):
         download_path = download_path.absolute()
         with chdir(directory_path):
-            shutil.unpack_archive(str(download_path), ".")
+            try:
+                shutil.unpack_archive(str(download_path), ".")
+            except:
+                shutil.rmtree(directory_path, onerror=on_error)
```
pdxjohnny, line 233: - I don't think we're using `onerror` in quite the way we should here. `rmtree`'s `onerror` is for when there are issues removing the extracted file.
  - We want to try to unpack the archive, if we fail, then we remote the directory, then we raise the error that occurred when unpacking. Or else the user won't know what happened. They'll be confused as to why we were trying and failing to delete their partially extracted archive.
  - It would be good to `except Exception as error`, then you can `raise DirectoryNotExtractedError(path) from error`.

- Can we get a test for this case? `unittest.mock` will be helpful here. There are examples throughout the tests (`git grep unittest.mock` or `git grep with\ patch`). You should be able to make `unpack_archive` raise an exception.

dffml/util/net.py
```diff
@@ -195,12 +211,26 @@ def cached_download_unpack_archive(
...
+    def on_error(func, path, exc_info):
+        """
+        If there's permission denied error. Error handler for shutil.rmtree
+        """
+        if not os.access(path, os.W_OK):
+            os.chmod(path, stat.S_IWUSR)
+            func(path)
+        else:
+            raise DirectoryNotExtractedError(path)
+
```
pdxjohnny, line 224: Per comment below

```suggestion
```

dffml/util/net.py
```diff
@@ -1,10 +1,13 @@
+import stat
+import os
 import shutil
 import hashlib
 import pathlib
 import functools
 import urllib.request
 from typing import List
 
+
```
pdxjohnny, line 10:
```suggestion
```

pdxjohnny (changes requested):

dffml/util/net.py
nitesh585: Actually! in windows, the common problem with rmtree is that it throws an exception of permission denied. So, to handle this I have added the onerror. But yeah! you are right. User may be confused about what is going on.
Now, I made the changes accordingly so that whenever we failed to unpack it raise an exception ```DirectoryNotExtractedError```.

dffml/util/net.py
nitesh585: Why we are removing this on_error method?

pdxjohnny:
Thanks for addressing those comments! Can we run `black` on this and get a testcase? Thanks!

nitesh585:
Sure! I'll work on testcase. Thanks :)

nitesh585:
@pdxjohnny  I have added the test for the failure of cached_download_unpack_archive.

pdxjohnny:
Well done! Thank you!

Issue #1007, opened 2021-02-21
docs: specifying the python version while installing the dffml module


aryanxk02:
When I install the DFFML module on my Ubuntu 20.04 using the command as specified in the documentation : 
`python -m pip install -U dffml`

I receive a message saying, 

`  command 'python3' from deb python3`
` command 'python' from deb python-is-python3`

So by writing the command as, `python3 -m pip install -U dffml` the installation begins. #

My current python version on my machine is `Python 3.8.5`

I think I can change this minor thing in the documentation. Kindly check and allow me. Thanks and Regards

mHash1m:
Hey @aryanxk02, I just tried recreating your issue and I didn't get the same issue. 
My guess is that your python alias is causing the issue. 
Try the following command  to check your default python version `python --version`.
Also try creating an alias by `alias python=python3.8` and then checking the version and installing the package.
You can also make the alias permanent by pasting the above command in your `~/.bashrc`

PR #1010, opened 2021-02-22
update command for installing dffml

aryanxk02:
command line changes

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -51,7 +51,7 @@ package.
 
         .. code-block:: console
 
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -51,7 +51,7 @@ package.
 
         .. code-block:: console
 
-            (.venv) C:\Users\username> python -m pip install -U dffml
+            (.venv) C:\Users\username> python3 -m pip install -U dffml
```
pdxjohnny, line 54: Is this related to issue #1007 ? If so, shouldn't we be changing line 48?

Issue #868, opened 2020-10-12
record: Remove confidence from prediction


pdxjohnny:
Some models use their accuracy as their confidence for a given prediction.

We should separate out the confidence for a prediction into a separate data structure so that we avoid using `float('nan')` when `ModelContext.predict` doesn't know the confidence.

pdxjohnny:
This is important because it make the access of prediction data uniform with the feature data.

`record.feature("a")` and ``record.prediction("b")` both give value when this is complete

PR #1020, opened 2021-02-27
record: Separate confidence from prediction

mHash1m:
Fixes: #868

### DIFF
--- a/dffml/model/slr.py
+++ b/dffml/model/slr.py
@@ -129,11 +129,11 @@ class SLRModel(SimpleModel):
                 },
                 "key": "0",
                 "last_updated": "2020-11-15T16:22:25Z",
-                "prediction": {
-                    "ans": {
-                        "confidence": 0.9355670103092784,
-                        "value": 1
-                    }
+                "predictions": {
+                    "ans": 1
+                }
+                "confidences": {
+                    "ans": 0.9355670103092784
                 }
             }
         ]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Calls to hashlib now go through helper functions
 - Build docs using `dffml service dev docs`
 - `cached_download/unpack_archive()` are now functions
+- Separate `confidences` from `predictions` and make it an optional parameter.
 ### Fixed
 - Record object key properties are now always strings
 
--- a/dffml/high_level.py
+++ b/dffml/high_level.py
@@ -274,7 +274,8 @@ async def save(source: BaseSource, *args: Record) -> None:
     ...             "myrecord",
     ...             data={
     ...                 "features": {"Years": 0, "Expertise": 1, "Trust": 0.1},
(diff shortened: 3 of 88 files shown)


dffml/model/slr.py
```diff
@@ -131,10 +131,14 @@ class SLRModel(SimpleModel):
                 "last_updated": "2020-11-15T16:22:25Z",
                 "prediction": {
                     "ans": {
-                        "confidence": 0.9355670103092784,
                         "value": 1
                     }
                 }
+                "confidence": {
+                    "ans": {
+                        "confidence": 0.9355670103092784,
```
pdxjohnny, line 139: I know I just posted the recording, but I want to confirm with you that the plan here is to have the following output when this PR is not WIP:

```
                "prediction": {
                    "ans": 1
                }
                "confidence": {
                    "ans": 0.9355670103092784
                }
```

pdxjohnny (commented):
Off to a great start. Just be sure to address the comment. Let me know if you have any questions.

dffml/model/slr.py
mHash1m: @pdxjohnny Ah! Tbh I thought we wanted to keep it so I made the whole PR around keeping the format :facepalm: . And I had actually completed the PR lol. But yeah this makes sense. I'll fix this asap.
Also about the recording, I made the csv work with the headers, or so it seems. Please review that part as well and let me know if we still need to have that property with predictions and confidences.

mHash1m: Also, we don't need the dicts inside `confidence` and `predictions` right? we could just store single values as in the above expected output you shared.

mHash1m: Also would we want to have the `record.predictions()` output something similar to whats stored inside records? As there are two different outputs, one of the actual record and the other of `record.predictions()` which also gets called from other files. So that would mean `record.predictions()` output:

> I know I just posted the recording, but I want to confirm with you that the plan here is to have the following output when this PR is not WIP:
> 
> ```
>                 "prediction": {
>                     "ans": 1
>                 }
>                 "confidence": {
>                     "ans": 0.9355670103092784
>                 }
> ```

pdxjohnny: > Also, we don't need the dicts inside `confidence` and `predictions` right? we could just store single values as in the above expected output you shared.

Yes

pdxjohnny: > Also would we want to have the `record.predictions()` output something similar to whats stored inside records? As there are two different outputs, one of the actual record and the other of `record.predictions()` which also gets called from other files. So that would mean `record.predictions()` output:
> 
> > I know I just posted the recording, but I want to confirm with you that the plan here is to have the following output when this PR is not WIP:
> > ```
> >                 "predictions": {
> >                     "ans": 1
> >                 }
> >                 "confidences": {
> >                     "ans": 0.9355670103092784
> >                 }
> > ```


Here's what it should look like from a REPL when the PR is over

```python
>>> record.predictions()
{'ans': 1}
```


```python
>>> record.confidences()
{'ans': 0.9355670103092784}
```

> >                 }
> >                 "confidences": {
> >                     "ans": 0.9355670103092784
> >                 }

pdxjohnny (changes requested):

Issue #909, opened 2020-11-24
model: pytorch: out_channels as in_channels


pdxjohnny:
out_channels of the previous layer should be used as in_channels of the next layer, unless overridden

SGeetansh:
May I work on this?

PR #1022, opened 2021-03-01
model: pytorch: out_channels as in_channels

Pratikrocks:
Signed-off-by: Pratik Dey <pratikrocks.dey11@gmail.com>

Fix for the issue #909

### DIFF
--- a/examples/rockpaperscissors/python_example.py
+++ b/examples/rockpaperscissors/python_example.py
@@ -18,10 +18,10 @@ class ConvNet(nn.Module):
             in_channels=3, out_channels=32, kernel_size=5, padding=2
         )
         self.conv2 = nn.Conv2d(
-            in_channels=32, out_channels=32, kernel_size=3, padding=1
+            in_channels=self.conv1.out_channels, out_channels=32, kernel_size=3, padding=1
         )
         self.conv3 = nn.Conv2d(
-            in_channels=32, out_channels=16, kernel_size=3, padding=1
+            in_channels=self.conv2.out_channels, out_channels=16, kernel_size=3, padding=1
         )
 
         self.relu = nn.ReLU()
--- a/examples/rockpaperscissors/python_example.py
+++ b/examples/rockpaperscissors/python_example.py
@@ -14,37 +14,71 @@ class ConvNet(nn.Module):
     def __init__(self, in_channels=3, num_features=3):
         super(ConvNet, self).__init__()
 
-        self.conv1 = nn.Conv2d(
-            in_channels=in_channels, out_channels=32, kernel_size=5, padding=2
+        self.layers = []
+        # BLOCK 1
+        self.layers.extend(
+            [
+                nn.Conv2d(
+                    in_channels=in_channels,
+                    out_channels=32,
+                    kernel_size=5,
+                    padding=2,
+                ),
+                nn.ReLU(inplace=True),
+                nn.MaxPool2d(kernel_size=2),
+            ]
         )
-        self.conv2 = nn.Conv2d(
-            in_channels=32, out_channels=32, kernel_size=3, padding=1
+        # BLOCK 2
+        self.layers.extend(
+            [
+                nn.Conv2d(
+                    in_channels=self.layers[0].out_channels,
+                    out_channels=32,
+                    kernel_size=3,
+                    padding=1,
+                ),
+                nn.ReLU(inplace=True),
+                nn.MaxPool2d(kernel_size=2),
+            ]
         )
-        self.conv3 = nn.Conv2d(
-            in_channels=32, out_channels=16, kernel_size=3, padding=1
+        # BLOCK 3
+        self.layers.extend(
+            [
+                nn.Conv2d(
+                    in_channels=self.layers[3].out_channels,
+                    out_channels=32,
+                    kernel_size=3,
+                    padding=1,
+                ),
+                nn.ReLU(inplace=True),
+                nn.MaxPool2d(kernel_size=2),
+            ]
+        )
+        # BLOCK 4
+        self.layers.extend(
+            [
+                nn.Conv2d(
+                    in_channels=self.layers[6].out_channels,
+                    out_channels=16,
+                    kernel_size=3,
+                    padding=1,
+                ),
+                nn.ReLU(inplace=True),
+                nn.MaxPool2d(kernel_size=2),
+            ]
         )
 
-        self.relu = nn.ReLU()
-        self.pooling = nn.MaxPool2d(kernel_size=2)
+        self.conv_block = nn.Sequential(*self.layers)
+        # output dimension of the conv layer is ((n+2p-k)/s+1)*((n+2p-k)/s+1)*Nc
+        # output features after passing through a MaxPool2d is ((n-k)/s + 1)*((n-k)/s + 1)*Nc
 
         self.linear = nn.Linear(
             in_features=16 * 9 * 9, out_features=num_features
         )
 
     def forward(self, x):
-        # block 1
-        x = self.pooling(self.relu(self.conv1(x)))
-
-        # block 2
-        x = self.pooling(self.relu(self.conv2(x)))
-
-        # block 3
-        x = self.pooling(self.relu(self.conv2(x)))
-
-        # block 4
-        x = self.pooling(self.relu(self.conv3(x)))
-
         # fully connected layer
+        x = self.conv_block(x)
         x = self.linear(x.view(-1, 16 * 9 * 9))
         return x
 
--- a/examples/rockpaperscissors/python_example.py
+++ b/examples/rockpaperscissors/python_example.py
@@ -11,23 +11,25 @@ class ConvNet(nn.Module):
     Convolutional Neural Network to classify hand gestures in an image as rock, paper or scissors
     """
 
-    def __init__(self):
+    def __init__(self, in_channels=3, num_features=3):
         super(ConvNet, self).__init__()
 
         self.conv1 = nn.Conv2d(
-            in_channels=3, out_channels=32, kernel_size=5, padding=2
+            in_channels=in_channels, out_channels=32, kernel_size=5, padding=2
         )
         self.conv2 = nn.Conv2d(
-            in_channels=self.conv1.out_channels, out_channels=32, kernel_size=3, padding=1
+            in_channels=32, out_channels=32, kernel_size=3, padding=1
         )
         self.conv3 = nn.Conv2d(
-            in_channels=self.conv2.out_channels, out_channels=16, kernel_size=3, padding=1
+            in_channels=32, out_channels=16, kernel_size=3, padding=1
         )
 
         self.relu = nn.ReLU()
         self.pooling = nn.MaxPool2d(kernel_size=2)
 
-        self.linear = nn.Linear(in_features=16 * 9 * 9, out_features=3)
+        self.linear = nn.Linear(
+            in_features=16 * 9 * 9, out_features=num_features
+        )
 
     def forward(self, x):
         # block 1
--- a/model/pytorch/dffml_model_pytorch/utils/utils.py
+++ b/model/pytorch/dffml_model_pytorch/utils/utils.py
@@ -17,7 +17,6 @@ def create_layer(layer_dict):
     with all the indented inner layers, else if the layer_dict is a single layer
     i.e. of type Dict[str, int/str/float/sequence/bool], that layer is returned.
     """
-
     sequential_dict = nn.Sequential()
     for name, layer in layer_dict.items():
         if isinstance(layer, dict):
@@ -32,6 +31,45 @@ def create_layer(layer_dict):
             layer_type = parameters.pop("layer_type")
             return getattr(nn, layer_type)(**parameters)
     return sequential_dict
+
+
+def create_network(network_dict):
+    sequential_network = nn.Sequential()
+    for name, layer in network_dict.items():
+        if isinstance(layer, dict):
+            parameters = {k: v for k, v in layer.items()}
+            layer_type = parameters.pop("layer_type")
+            try:
+                if not isinstance(parameters["in_channels"], int):
+                    layers = []
+                    for module in sequential_network.children():
+                        layers.append(module)
+                    for module in reversed(layers):
+                        try:
+                            parameters["in_channels"] = module.out_channels
+                            break
+                        except:
+                            pass
+            except KeyError:
+                pass
+
+            try:
+                if not isinstance(parameters["in_features"], int):
+                    layers = []
+                    for module in sequential_network.children():
+                        layers.append(module)
+                    for module in reversed(layers):
+                        try:
+                            parameters["in_features"] = module.out_features
+                            break
+                        except:
+                            pass
+            except KeyError:
+                pass
+            sequential_network.add_module(
+                name, getattr(nn, layer_type)(**parameters)
+            )
+    return sequential_network
 
 
 class NumpyToTensor(Dataset):
--- a/model/pytorch/tests/test_pytorchnet.py
+++ b/model/pytorch/tests/test_pytorchnet.py
@@ -16,27 +16,72 @@ class ConvNet(nn.Module):
     def __init__(self, in_channels=3, num_features=3):
         super(ConvNet, self).__init__()
 
-        self.conv1 = nn.Conv2d(
-            in_channels=in_channels, out_channels=32, kernel_size=5, padding=2
+        self.layers = []
+        # BLOCK 1
+        self.layers.extend(
+            [
+                nn.Conv2d(
+                    in_channels=in_channels,
+                    out_channels=32,
+                    kernel_size=5,
+                    padding=2,
+                ),
+                nn.ReLU(inplace=True),
+                nn.MaxPool2d(kernel_size=2),
+            ]
         )
-        self.conv2 = nn.Conv2d(
-            in_channels=32, out_channels=32, kernel_size=3, padding=1
+        # BLOCK 2
+        self.layers.extend(
+            [
+                nn.Conv2d(
+                    in_channels=self.layers[0].out_channels,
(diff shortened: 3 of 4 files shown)


sakshamarora1:
@Pratikrocks I think you misunderstood the issue, we should be checking if the previous layer has an `out_channels/out_features` parameter and the next layer might have an `in_channels/in_features` parameter which in case is not specified by the user then should be auto assigned as the value from the last layer's `out_channels/out_features` (if it is a valid parameter for that layer).

Pratikrocks:
@sakshamarora1  yes you are correct I misunderstood the issue.
Since the parameters are hardcoded here(in the `conv nets`) so basically as per the issue we should modify it to an extent that it works for the random user-given input , and in case it is not set we would be using the prev layers `out_channels`. Right?

sakshamarora1:
Yes! For those layer functions that have those in/out parameters.

Pratikrocks:
@sakshamarora1  I think now it mostly solves the issue , and also in future we are to expand this net it would be a bit easier to just extend the `layers`

sakshamarora1:
@Pratikrocks The changes you have made are in an example on how dffml's pytorch model plugin can used via the python API. The issue points to adding a functionality inside plugin's code base, the functionality being what I explained above in my previous comment, when a network is defined using the config file.

Reference:
https://github.com/intel/dffml/blob/91172cfe4e672886377139cb59a33fdb4cbbf0b3/model/pytorch/dffml_model_pytorch/pytorch_net.py#L14-L15
https://github.com/intel/dffml/blob/91172cfe4e672886377139cb59a33fdb4cbbf0b3/model/pytorch/dffml_model_pytorch/utils/utils.py#L14-L19

Pratikrocks:
Okay @sakshamarora1 looking at it shortly

Pratikrocks:
@sakshamarora1 kindly review the changes once weather its going as per the issue requirement :)

Pratikrocks:
@pdxjohnny @sakshamarora1 can you please review the changes once :)

model/pytorch/tests/test_pytorchnet.py
```diff
@@ -183,6 +184,57 @@ async def test_02_predict(self, predictdir):
...
+            "conv2": {
+                "layer_type": "Conv2d",
+                "in_channels": None,
+                "out_channels": 64,
+                "kernel_size": 3,
+            },
+            "relu2": {"layer_type": "ReLU"},
+            "conv3": {
+                "layer_type": "Conv2d",
+                "in_channels": "",
```
sakshamarora1, line 218: We should be able to make it such that we don't need to mention it at all
```suggestion
```

model/pytorch/dffml_model_pytorch/utils/utils.py
```diff
@@ -34,6 +33,45 @@ def create_layer(layer_dict):
...
+                    for module in reversed(layers):
+                        try:
+                            parameters["in_channels"] = module.out_channels
+                            break
+                        except:
+                            pass
+            except KeyError:
+                pass
+
+            try:
```
sakshamarora1, line 56: I see that you have used multiple try except blocks. What are all the cases they are handling?

model/pytorch/dffml_model_pytorch/utils/utils.py
```diff
@@ -34,6 +33,45 @@ def create_layer(layer_dict):
     return sequential_dict
 
 
+def create_network(network_dict):
```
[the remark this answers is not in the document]
Pratikrocks: This function is basically taking an dict(in the form of https://github.com/intel/dffml/blob/master/model/pytorch/examples/resnet18/layers.yaml) which contains a model to be trained , and then its converting to a `Sequential model` which will eventually be converted as a `PytorchNeuralNetwork` [like we have this now](https://github.com/intel/dffml/blob/42bb2aa101562ca6dbaeded6acf1125edbdcfe27/model/pytorch/tests/test_pytorchnet.py#L65). Moreover I saw that presently if we are passing a dict of the models instead of `nn.Sequential` we were getting an error from the `test cases`

model/pytorch/dffml_model_pytorch/utils/utils.py
```diff
@@ -34,6 +33,45 @@ def create_layer(layer_dict):
...
+    sequential_network = nn.Sequential()
+    for name, layer in network_dict.items():
+        if isinstance(layer, dict):
```
Pratikrocks, line 39: We are checking for a dict object it is so then continue

Pratikrocks: `model2 = {
            "conv1": {
                "layer_type": "Conv2d",
                "in_channels": 3,
                "out_channels": 32,
                "kernel_size": 3,
            },
            "relu1": {"layer_type": "ReLU"},
            "conv2": {
                "layer_type": "Conv2d",
                "in_channels": None,
                "out_channels": 64,
                "kernel_size": 3,
            },
            "relu2": {"layer_type": "ReLU"},
            "conv3": {
                "layer_type": "Conv2d",
                "in_channels": "",
                "out_channels": 96,
                "kernel_size": 3,
            },
            "relu3": {"layer_type": "ReLU"},
            "conv4": {
                "layer_type": "Conv2d",
                "in_channels": "96",
                "out_channels": 32,
                "kernel_size": 3,
            },
        }`

Basicall this function is taking up this kinda models as its parameters

Pratikrocks: In `conv3` the `in_channels` are not set but it is valid as the `conv` layer will be taking its `in_channel` param from the `out_channel` of the previous `conv` layers, so this function is basically handling those stuffs and its converting it to a valid `nn.Sequential`  model

model/pytorch/dffml_model_pytorch/utils/utils.py
```diff
@@ -34,6 +33,45 @@ def create_layer(layer_dict):
...
+            parameters = {k: v for k, v in layer.items()}
+            layer_type = parameters.pop("layer_type")
```
Pratikrocks, line 41: for each `layer_type` we are popping its 'label' and later assigning it to a valid `nn` model

model/pytorch/dffml_model_pytorch/utils/utils.py
```diff
@@ -34,6 +33,45 @@ def create_layer(layer_dict):
...
+            try:
+                if not isinstance(parameters["in_channels"], int):
```
Pratikrocks, line 43: `in_channels ` may or may not be present so we are handling it in `try` and `except` (line 53) block , In except we need not do any thing as `in_channels` may be absent like `ReLU, MaxPool ....`

model/pytorch/dffml_model_pytorch/utils/utils.py
```diff
@@ -34,6 +33,45 @@ def create_layer(layer_dict):
...
+                    layers = []
+                    for module in sequential_network.children():
+                        layers.append(module)
+                    for module in reversed(layers):
```
Pratikrocks, line 47: Here if the looking at the children of the `nn.Sequential` from reverse as the `in_channels` will be assigned a value which is just above it

model/pytorch/dffml_model_pytorch/utils/utils.py
```diff
@@ -34,6 +33,45 @@ def create_layer(layer_dict):
...
+                        try:
+                            parameters["in_channels"] = module.out_channels
+                            break
```
Pratikrocks, line 50: In this case if there is a `out_children` then break,  else further continue for other layers

model/pytorch/dffml_model_pytorch/utils/utils.py
Pratikrocks: Explained above

model/pytorch/dffml_model_pytorch/utils/utils.py
```diff
@@ -34,6 +33,45 @@ def create_layer(layer_dict):
...
+                        except:
+                            pass
+            except KeyError:
+                pass
+
+            try:
+                if not isinstance(parameters["in_features"], int):
+                    layers = []
```
Pratikrocks, line 58: The above was for `Conv model` this is for the `in/out_features` in the `nn.Linear` layers.

model/pytorch/tests/test_pytorchnet.py
Pratikrocks: Yes the function which I did actually handles it , as this input is valid, and the `conv3` layer will be taking its value from the  above `conv2` layers

Pratikrocks:
@sakshamarora1 one more thing, just out of curoisity,  is this [line](https://github.com/intel/dffml/blob/42bb2aa101562ca6dbaeded6acf1125edbdcfe27/model/pytorch/dffml_model_pytorch/pytorch_net.py#L17) necessary,  as `__init__` is required to return `None` always [ref](https://stackoverflow.com/questions/2491819/how-to-return-a-value-from-init-in-python#:~:text=10%20Answers&text=__init__%20is%20required,instance%20variable%20(or%20function).)

examples/rockpaperscissors/python_example.py
```diff
@@ -11,38 +11,74 @@ class ConvNet(nn.Module):
     Convolutional Neural Network to classify hand gestures in an image as rock, paper or scissors
     """
 
-    def __init__(self):
+    def __init__(self, in_channels=3, num_features=3):
```
sakshamarora1, line 14: This file shouldn't be changed as it's an example file.

model/pytorch/dffml_model_pytorch/utils/utils.py
sakshamarora1: I think the if else statements can handle that. We shouldn't need try except blocks for that.

model/pytorch/dffml_model_pytorch/utils/utils.py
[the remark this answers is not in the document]
sakshamarora1: Also, instead of creating a completely new function, we should be updating the already existing function `create_layer`

model/pytorch/tests/test_pytorchnet.py
```diff
@@ -13,30 +14,75 @@
 
 
 class ConvNet(nn.Module):
-    def __init__(self):
+    def __init__(self, in_channels=3, num_features=3):
         super(ConvNet, self).__init__()
 
-        self.conv1 = nn.Conv2d(
-            in_channels=3, out_channels=32, kernel_size=5, padding=2
+        self.layers = []
```
sakshamarora1, line 20: Again, all these changes in this class are just a different way of defining the layers. I don't think they are related to the issue (The function `create_layer` is not called in this case as it is already a PyTorch Module)

model/pytorch/dffml_model_pytorch/utils/utils.py
sakshamarora1: We should refrain from using the try except blocks, and look for ways to handle all the cases without them. (As it makes the code less readable and might bypass errors that shouldn't go unnoticed in the future)

model/pytorch/dffml_model_pytorch/utils/utils.py
sakshamarora1: [Reference](https://intel.github.io/dffml/master/tutorials/neuralnetwork.html#command-line), a sequential layer is only created when the layers are indented inside a key name like `block1:`, else the layers are added normally (in the model.yaml file you can see how each conv layer is called explicitly in the forward method)

There is a case when there is no sequential model and the layers are feeded to the `create_layer` function one by one. Maybe we should pass the complete model dict in the `pytorch_base.py` file instead of passing layers one by one.

sakshamarora1 (changes requested):
@Pratikrocks in the case of `__init__` , we are first checking if the network passed by the user is already a PyTorch nn.module (when defined using the Python API, in `python_example.py` file), if it is, then we are just returning it as it is, else if it is a dict, then we are creating the Network object.

examples/rockpaperscissors/python_example.py
Pratikrocks: Anyways having a slight variation , I think it won't affect the generality , will it ?

model/pytorch/dffml_model_pytorch/utils/utils.py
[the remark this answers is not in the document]
Pratikrocks: @sakshamarora1 just wanted to be sure , actually I was getting an error when I was passing a dict model to that class (In test currently we are having only the tests with `nn` models), have you also also faced the same error , or is it an issue from my side only?

model/pytorch/dffml_model_pytorch/utils/utils.py
Pratikrocks: Yes but if we pass only the layers one by one ,  we may loose the track of the `out_channels` of the above layers(but it's required as per the issue) .So I thaught of passing the entire `dict`

model/pytorch/dffml_model_pytorch/utils/utils.py
Pratikrocks: okay I will change it

model/pytorch/tests/test_pytorchnet.py
Pratikrocks: Yeah its not related ,  I will revert this particular section

Pratikrocks:
@sakshamarora1 can you please verify the following 
        model2 = {
            "conv1": {
                "layer_type": "Conv2d",
                "in_channels": 3,
                "out_channels": 32,
                "kernel_size": 3,
            },
            "relu1": {"layer_type": "ReLU"},
            "conv2": {
                "layer_type": "Conv2d",
                "in_channels": None,
                "out_channels": 64,
                "kernel_size": 3,
            },
            "relu2": {"layer_type": "ReLU"},
            "conv3": {
                "layer_type": "Conv2d",
                "in_channels": "",
                "out_channels": 96,
                "kernel_size": 3,
            },
            "relu3": {"layer_type": "ReLU"},
            "conv4": {
                "layer_type": "Conv2d",
                "in_channels": "96",
                "out_channels": 32,
                "kernel_size": 3,
            },
        }

        model = PyTorchNeuralNetwork(
            classifications=["rock", "paper", "scissors"],
            features=Features(Feature("image", int, 300 * 300)),
            predict=Feature("label", int, 1),
            directory='/home/pratik/dffml',
            network=model2
        )
        print(model)

Pratikrocks:
The unit are eventually failing presently on passing a `dict` object
`
ERROR: test_00_train (tests.test_pytorchnet.TestPyTorchNeuralNetwork)
  File "/home/pratik/dffml/dffml/util/asynctestcase.py", line 69, in run_it
  File "/usr/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "/home/pratik/dffml/dffml/util/net.py", line 215, in wrapper
    return await func(*(list(args) + [directory_path]), **kwds)
  File "/home/pratik/dffml/model/pytorch/tests/test_pytorchnet.py", line 165, in test_00_train
    await train(
  File "/home/pratik/dffml/dffml/high_level.py", line 404, in train
    mctx = await astack.enter_async_context(model())
  File "/usr/lib/python3.8/contextlib.py", line 568, in enter_async_context
    result = await _cm_type.__aenter__(cm)
  File "/home/pratik/dffml/model/pytorch/dffml_model_pytorch/pytorch_base.py", line 94, in __aenter__
    self._model = self.createModel()
  File "/home/pratik/dffml/model/pytorch/dffml_model_pytorch/pytorch_net.py", line 80, in createModel
    self._model = model.to(self.device)
AttributeError: 'dict' object has no attribute 'to'

Ran 1 test in 146.015s
`

examples/rockpaperscissors/python_example.py
sakshamarora1: No, it won't, but changing a working and clear code is basically redundant.

Pratikrocks: Okay I will revert these changes

sakshamarora1:
@Pratikrocks It is working for me. Cleaning the branch to only have the required changes might resolve it.

Pratikrocks:
Hey @sakshamarora1 can you post me the output of the console of what you are receiving after running this test

```
async def test_03_network_creation_from_dict(self):
        model1 = {
            "linear1": {
                "layer_type": "Linear",
                "in_features": 4096,
                "out_features": 256,
            },
            "relu": {"layer_type": "ReLU"},
            "linear_2": {
                "layer_type": "Linear",
                "in_features": "",
                "out_features": 17,
            },
        }
        model2 = {
            "conv1": {
                "layer_type": "Conv2d",
                "in_channels": 3,
                "out_channels": 32,
                "kernel_size": 3,
            },
            "relu1": {"layer_type": "ReLU"},
            "conv2": {
                "layer_type": "Conv2d",
                "in_channels": None,
                "out_channels": 64,
                "kernel_size": 3,
            },
            "relu2": {"layer_type": "ReLU"},
            "conv3": {
                "layer_type": "Conv2d",
                "in_channels": "",
                "out_channels": 96,
                "kernel_size": 3,
            },
            "relu3": {"layer_type": "ReLU"},
            "conv4": {
                "layer_type": "Conv2d",
                "in_channels": "96",
                "out_channels": 32,
                "kernel_size": 3,
            },
        }

        model1 = PyTorchNeuralNetwork(
            classifications=["rock", "paper", "scissors"],
            features=Features(Feature("image", int, 300 * 300)),
            predict=Feature("label", int, 1),
            directory='/',
            network=model1,
        )

        await train(
            model1,
            DirectorySource(
                foldername= "/",
                feature="image",
                labels=["rock", "paper", "scissors"],
            ),
        )
```

From my side, I am receiving the error as it's unable to convert to the `nn` layers
You can just append this extra test function in `
model/pytorch/tests/test_pytorchnet.py`   
and just run the specific test using   `python3 setup.py test -s tests.test_pytorchnet.TestPyTorchNeuralNetwork.test_03_network_creation_from_dict`

PR #1023, opened 2021-03-01
util: crypto: Create secure_hash and insecure_hash functions

### LINKED ISSUE #297: util: crypto: Create secure_hash and insecure_hash functions
Throughout the codebase we have calls to `hashlib`, usually `md5`, `sha256`, or `sha384`.

It would be good to have the calls to hashlib go through a helper function. So we can better track where hashes are being used for convenience (to shorten an otherwise long value), or where hashes are being used there might be a security consequnce to using the wrong hash.

Change to `insecure_hash`
- `md5` is currently only used in the dataflow visualization.
- In the model saving and loading code (`_filename` and similar)

Use `secure_hash`
- Everywhere in df/memory.py (until further notice)
- Everywhere else

kahanikaar:
Fixes #297

### DIFF
--- a/configloader/image/tests/test_config.py
+++ b/configloader/image/tests/test_config.py
@@ -1,8 +1,8 @@
 import json
 import pathlib
-import hashlib
 
 from dffml.util.asynctestcase import AsyncTestCase
+from dffml.util.crypto import secure_hash
 from dffml_config_image.configloader import PNGConfigLoader
 
 IMAGE1_HASH = "6faf9050c6d387bc6a68d9e12127f883011add2ec994b8e66c7c0996636f2789af8d28fc11e6528a327a6383c1473e72"
@@ -22,8 +22,8 @@ class TestConfig(AsyncTestCase):
                     / "image1.png"
                 ).read_bytes()
                 original = await ctx.loadb(image_bytes)
-                hash_original = hashlib.sha384(
-                    json.dumps(original.flatten().tolist()).encode()
-                ).hexdigest()
+                hash_original = secure_hash(
+                    json.dumps(original.flatten().tolist()), algorithm="sha384"
+                )
                 self.assertEqual(original.shape, (280, 280, 3))
                 self.assertEqual(hash_original, IMAGE1_HASH)
--- a/model/scikit/dffml_model_scikit/scikit_base.py
+++ b/model/scikit/dffml_model_scikit/scikit_base.py
@@ -4,7 +4,6 @@
 Base class for Scikit models
 """
 import json
-import hashlib
 import pathlib
 import logging
 import importlib
@@ -30,6 +29,7 @@ from dffml.model.accuracy import Accuracy
 from dffml.source.source import Sources, SourcesContext
 from dffml.model.model import ModelConfig, ModelContext, Model, ModelNotTrained
 from dffml.feature.feature import Features, Feature
+from dffml.util.crypto import secure_hash
 
 
 class ScikitConfig(ModelConfig, NamedTuple):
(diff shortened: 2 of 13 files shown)


configloader/image/tests/test_config.py
```diff
@@ -1,8 +1,8 @@
 import json
 import pathlib
-import hashlib
 
 from dffml.util.asynctestcase import AsyncTestCase
+from dffml.util.net import secure_hash
```
pdxjohnny, line 5: `util: crypto` in the issue description means that `*_hash` functions should go in `dffml/util/crypto.py`

dffml/util/net.py
```diff
@@ -8,6 +8,23 @@
...
+    if encoding == "yes":
+        string = string.encode("utf-8")
+    if algorithm == "sha256":
+        return hashlib.sha256(string).hexdigest()
+    elif algorithm == "sha384":
+        return hashlib.sha384(string).hexdigest()
+
+
+def insecure_hash(string, algorithm="md5", encoding="yes") -> str:
+    if encoding == "yes":
```
pdxjohnny, line 22: In the case that the caller passes bytes, we don't want to try to re-encode.

Also, is `encoding` currently used by any callers? Would it make more sense to just have an `isinstance()` check?

```suggestion
    if encoding == "yes" and isinstance(string, str):
```

dffml/util/net.py
```diff
@@ -8,6 +8,23 @@
...
+        string = string.encode("utf-8")
+    if algorithm == "md5":
+        return hashlib.md5(string).hexdigest()
```
pdxjohnny, line 25: We shouldn't have a case where we return None here, the caller wouldn't be expecting that. We should either `raise` an exception if algorithm is not md5, or we should always use md5, which I'd say is a safe assumption for now. A safe assumption since you've made algorithm a keyword argument which defaults to md5, if you're worried about forward compatibility, you really aren't gaining anything by adding this keyword argument now. One could always add it as a keyword argument later and add more functionality, so it probably is best for us to just always return the md5 hash

dffml/util/net.py
```diff
@@ -111,7 +128,9 @@ def cached_download(
     target_path = pathlib.Path(target_path)
 
     def validate_hash(error: bool = True):
-        filehash = hashlib.sha384(target_path.read_bytes()).hexdigest()
+        filehash = secure_hash(
+            target_path.read_bytes(), algorithm="sha384", encoding="no"
```
pdxjohnny, line 132: This is a place where if we went with the `instance()` approach in `secure_hash()`, we wouldn't need to explictly set the encoding keyword argument here

model/scikit/dffml_model_scikit/scikit_base.py
```diff
@@ -64,9 +64,9 @@ def _feature_predict_hash(self):
                 if k not in ["features", "tcluster", "predict"]
             ]
         )
-        return hashlib.sha384(
-            "".join([params] + self.features).encode()
-        ).hexdigest()
+        return secure_hash(
+            "".join([params] + self.features), algorithm="sha384"
```
pdxjohnny, line 68: The same situation with what we discussed with `insecure_hash(algorithm="md5")`

pdxjohnny (changes requested):
Off to a great start!

configloader/image/tests/test_config.py
kahanikaar: Fixed that! hash helper functions are now  in dffml/util/crypto.py

dffml/util/net.py
kahanikaar: Encoding argument removed, now isinstance is checked.

dffml/util/net.py
kahanikaar: insecure_hash now always uses md5 hash.

dffml/util/net.py
kahanikaar: Encoding argument removed, isinstance is added.

pdxjohnny:
I added a test to check were hashlib is used. This will hopefully prevent people from using it instead of the functions you added. Since if they see that the test fails they'll ether have to add to the `dict` of audited, and we'd catch it during review and tell them to use your new functions. Thank you! Great work.

PR #1043, opened 2021-03-15
source: csv: Strip spaces in columns by default

### LINKED ISSUE #317: source: csv: Strip spaces in columns by default
`CSVSource` does not strip spaces on the right side of text while reading, this errors out when we have space after the last column name.

Pretty sure spaces being exported are a formatting error with whatever does the export to CSV.

We should probably add another property to the `CSVSoruceConfig` which tells it to strip spaces by default unless the flag is given. Something like `nostrip`. Since most of the time people will want it to strip spaces sliently to get around those pesky bad exporters.

Reporter: @0dust

nitesh585:
fix #317

### DIFF
--- a/dffml/source/csv.py
+++ b/dffml/source/csv.py
@@ -43,6 +43,7 @@ CSV_SOURCE_CONFIG_DEFAULT_KEY = "key"
 CSV_SOURCE_CONFIG_DEFAULT_tag = "untagged"
 CSV_SOURCE_CONFIG_DEFAULT_tag_COLUMN = "tag"
 CSV_SOURCE_CONFIG_DEFAULT_LOADFILES_NAME = None
+CSV_SOURCE_CONFIG_DEFAULT_NOSTRIP = False
 
 
 @config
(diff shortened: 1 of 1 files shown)


dffml/source/csv.py
```diff
@@ -100,7 +102,12 @@ async def read_csv(self, fd, open_file):
         open_file.write_out = {}
         # If there is no key track row index to be used as key by tag
         index = {}
+
         for row in dict_reader:
+            # strip the keys and values of row
+            if not self.config.nostrip:
+                row = {k.strip(): v.strip() for k, v in row.items()}
+
```
pdxjohnny, line 110:
```suggestion
```

dffml/source/csv.py
```diff
@@ -100,7 +102,12 @@ async def read_csv(self, fd, open_file):
         open_file.write_out = {}
         # If there is no key track row index to be used as key by tag
         index = {}
+
         for row in dict_reader:
+            # strip the keys and values of row
```
pdxjohnny, line 107:
```suggestion
            # Strip the keys and values of row
```

dffml/source/csv.py
```diff
@@ -100,7 +102,12 @@ async def read_csv(self, fd, open_file):
         open_file.write_out = {}
         # If there is no key track row index to be used as key by tag
         index = {}
+
```
pdxjohnny, line 105:
```suggestion
```

PR #1048, opened 2021-03-20
WIP:source:dataframe:Implements pandas.DataFrame

### LINKED ISSUE #925: source: pandas: Take pandas DataFrame
Fixing this issue will improve the user friendliness of DFFML. Users will be able to pass pandas DataFrame objects to the high level functions when this issue is completed. We need to implement a source that acts as a proxy (passthrough) between DFFML's APIs and a pandas DataFrame. This will let users pass DataFrames to high level functions without converting them to records first.

The `records()` and `record()` methods are working in this branch: https://github.com/pdxjohnny/dffml/commits/source_dataframe

The update method and the tests are still needed.

The changes also include a modification to the high level interface to instantiate an instance of the source if we detect that one of the high level sources has been passed a [`pandas.DataFrame`](https://pandas.pydata.org/pandas-docs/stable/reference/frame.html) in an argument.

spur19:
Aims to resolve #925


Pratikrocks:
Hi @spur19 I thi as per the issue we also need to modify the File source class base, so that it accepts Pandas dataframe too apart from records , and then further to high level fn , right ?

pdxjohnny:
All that needs to happen is to take https://github.com/pdxjohnny/dffml/commit/c188b66bd6b03d6a51dce5f1c6f1d08e34d5b51b and add the `.update()` method and some tests. I see that this PR is WIP, but just wanted to reiterate that everything else should be removed.

pdxjohnny:
Any progress on this?

Issue #1068, opened 2021-04-04
Implementation of k-fold technique


Saanidhyavats:
## Pain Point

K-fold technique method is more robust as compared to other methods for measuring the accuracy of a model. For implementing the idea of Automl, this feature will be beneficial.

## Proposed Solution

What would make things behave most smoothly

Saanidhyavats:
@pdxjohnny I am thinking to work on this. Shall I implement it from scratch?

mHash1m:
Will the conclusion of this issue include something like [`cross_val_score`](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.cross_val_score.html#sklearn-model-selection-cross-val-score)? Maybe returning accuracy for all the folds?

PR #1090, opened 2021-04-10
Added a function to split the data into n folds

Saanidhyavats:
This pr solves a part of issue #1068 . Will be creating another pr to solve the other parts.

### DIFF
--- a/tests/model/test_kfold.py
+++ b/tests/model/test_kfold.py
@@ -1,28 +1,28 @@
 import unittest
+import asyncio
+import pathlib
 from random import randrange
 
 from dffml import *
 
 
 class Kfold:
-    async def split(
-        self, source, fold_size, shuffle=False, *args
-    ):  # Will split the data into required number of folds
-        fold = list()
-        split_data = list()
-        async for record in load(source):
-            if len(fold) < fold_size:
-                fold.append(record.export())
-            else:
-                split_data.append(fold)
-                fold = []
-        return split_data
-
+	async def split(
+		self, source, fold_size, *args
+	):  # Will split the data into required number of folds
+		fold = list()
+		split_data = list()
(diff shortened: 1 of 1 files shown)


Saanidhyavats:
@pdxjohnny I have created a file as discussed in the meeting. Let me know if any change needs to be done.

tests/model/test_kfold.py
```diff
@@ -0,0 +1,31 @@
...
+        for i in range(folds):
+            x = i * fold_size
+            fold = list()
+            for j in range(x, x + fold_size):
+                fold.append(data[j])
+            split_data.append(fold)
+        return split_data
+
+
+class TestKfoldTechnique:
```
pdxjohnny, line 25:
```suggestion
class TestKfoldTechnique(unittest.TestCase):
```

pdxjohnny: This will ensure the code gets tested

tests/model/test_kfold.py
```diff
@@ -0,0 +1,31 @@
+from random import randrange
+
+import pandas as pd
```
pdxjohnny, line 3: We can't use any third party dependencies within the main `dffml` package

tests/model/test_kfold.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class Kfold:
+    def split(
```
pdxjohnny, line 7: We need to make sure that this works when streaming data. Sources have an `async` `records()` method which yields one record at a time. We have to make sure any method we use doesn't require the whole dataset be in memory at the same time

Saanidhyavats:
@pdxjohnny I have added the test for split method. Currently working on accuracy part.

Issue #1056, opened 2021-03-31
util : net : `cached_download` and derivatives should no longer be decorators  


programmer290399:
* Use of `cached_download` and derivatives is difficult in certain situations  because they work as decorators.
     - For ex: We want to download and validate a file inside a function's body, so it may not be possible to pass the parameters required by `cached_download` when it is being used as a decorator as these parameters may be generated inside the function's body. 
* Cached download  can be and technically is a synchronous operation 
* Thus these functions no longer need to be decorators.
* Relevant Discussion from [Weekly Sync 2021-03-30](https://youtu.be/LQfoxbO_u3A?t=2300)

PR #1092, opened 2021-04-11
util: net: Change cached_download and derivatives to functions.

mHash1m:
Fixes: #1056

### DIFF
--- a/dffml/util/net.py
+++ b/dffml/util/net.py
@@ -154,7 +154,7 @@ def sync_urlretrieve_and_validate(
     validate_file_hash(
         target_path, expected_sha384_hash=expected_sha384_hash,
     )
-    return str(target_path.absolute())
+    return target_path.absolute()
 
 
 async def cached_download(
--- a/dffml/util/net.py
+++ b/dffml/util/net.py
@@ -8,7 +8,6 @@ import contextlib
 import dataclasses
 import urllib.request
 from typing import List, Union
-
 from .os import chdir
 from .file import validate_file_hash
 from .log import LOGGER, get_download_logger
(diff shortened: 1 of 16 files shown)


dffml/util/net.py
```diff
@@ -406,17 +328,17 @@ def cached_download_unpack_archive(
...
     ...     "dffml",
     ...     "bb9bb47c4e6e4c6b7147bb3c000bc4069d69c0c77a3e560b69f476a78e6b5084adf5467ee83cbbcc47ba5a4a0696fdfc",
-    ... )
-    ... async def files_in_dffml_commit_c4469a(dffml_dir):
+    ...     )
     ...     return len(list(dffml_dir.rglob("**/*")))
     >>>
-    >>> asyncio.run(files_in_dffml_commit_c4469a())
-    124
+    >>> print(asyncio.run(files_in_dffml_commit_c4469a()))
```
pdxjohnny, line 340: Same here, we can move the call to `cached_download_unpack_archive()` into the top level of the example, call with `asyncio.run()`, and then get rid of the function, just read the directory contents from the top level.

examples/shouldi/tests/binaries.py
```diff
@@ -3,77 +3,101 @@
...
 downloads = pathlib.Path(__file__).parent / "downloads"
 
+loop = asyncio.get_event_loop()
 # JavaScript
-cached_node = cached_download_unpack_archive(
-    "https://nodejs.org/dist/v14.2.0/node-v14.2.0-linux-x64.tar.xz",
-    downloads / "node.tar.gz",
-    downloads / "node-download",
-    "fa2a9dfa4d0f99a0cc3ee6691518c026887677a0d565b12ebdcf9d78341db2066427c9970c41cbf72776a370bbb42729",
+cached_node = loop.run_until_complete(
```
pdxjohnny, line 8: We had talk at one point (a while ago) about how this file is referenced in the workflow. If the file changes, the cache directory of `downloads` is invalidated. With the way you have it now, these downloads run on import, we need to make it so that the arguments are stored in this file, and the functions they previously decorated call the download function using the arguments stored in this file.

model/pytorch/tests/test_pytorchnet.py
```diff
@@ -81,12 +82,34 @@ def setUpClass(cls):
...
         cls.model_dir.cleanup()
 
-    @cached_download_unpack_archive(
-        "https://storage.googleapis.com/laurencemoroney-blog.appspot.com/rps.zip",
-        "rps.zip",
-        "traindir",
-        "c6a9119b0c6a0907b782bd99e04ce09a0924c0895df6a26bc6fb06baca4526f55e51f7156cceb4791cc65632d66085e8",
+    loop = asyncio.get_event_loop()
+
+    traindir = loop.run_until_complete(
```
pdxjohnny, line 87: Same thing, why not use `asyncio.run()`?

pdxjohnny (changes requested):
Nice! Just a few minor things. Thanks!

examples/shouldi/tests/binaries.py
mHash1m: @pdxjohnny I've made some changes and I hope that's what was expected. However, the shouldi tests broke and I'm not sure why as the logs are ambiguous. The following error pops up in the logs:
```
ERROR: test_run (tests.test_dependency_check.TestRunDependencyCheckOp)
  File "/home/runner/work/dffml/dffml/dffml/util/asynctestcase.py", line 69, in run_it
  File "/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/asyncio/base_events.py", line 587, in run_until_complete
    return future.result()
  File "/home/runner/work/dffml/dffml/examples/shouldi/tests/test_dependency_check.py", line 35, in test_run
    results = await run_dependency_check(str(rxjava / "RxJava-2.2.16"))
  File "/home/runner/work/dffml/dffml/examples/shouldi/shouldi/java/dependency_check.py", line 66, in run_dependency_check
    raise DependencyCheckError(stderr.decode())
shouldi.java.dependency_check.DependencyCheckError

```

Tried reviewing my own code but found nothing as the error doesn't say much about what is at fault. 
Made the following debug statements in `run_dependency_check`:
```
                logging.debug(stderr)
                logging.debug(pkg)
                logging.debug(proc.returncode)
                logging.debug(stderr.decode())
```
which gave the following in logs:
```
DEBUG:root:b''
DEBUG:root:/home/runner/work/dffml/dffml/examples/shouldi/tests/downloads/RxJava-download/RxJava-2.2.16
DEBUG:root:243
DEBUG:root:
ERROR
```
note that `stderr.decode()` gives a blank line.

pdxjohnny (commented):
Couple more things, I'm not sure what's wrong with the java stuff, let's revert the logging changes for now.

dffml/util/net.py
```diff
@@ -432,26 +287,20 @@ def on_error(func, path, exc_info):
...
-            return await func(*(list(args) + [directory_path]), **kwds)
-
-        return wrapper
-
-    return mkwrapper
+        try:
+            shutil.unpack_archive(str(download_path), str(directory_path))
+        except Exception as error:
+            shutil.rmtree(directory_path, onerror=on_error)
+            raise DirectoryNotExtractedError(directory_path) from error
```
mHash1m, line 294: took me solid hours to figure this one out.  Apparently ` shutil.unpack_archive()` doesn't recognize "__.zip" as a zip file when using `with chdir(directory_path):`, I suppose we're returning the path to the zip file directly instead of the directory containing it so that could be causing it, so I just passed the path directly to ` shutil.unpack_archive()`.

mHash1m: The test was weirdly passing when I didn't use Asynio.run() so it had me running in circles around using loops and `asyncio.run()`.

dffml/util/net.py
```diff
@@ -432,26 +287,20 @@ def on_error(func, path, exc_info):
...
+        directory_path.mkdir(parents=True)
+        await extractor(
+            await cached_download(
+                url,
+                file_path,
+                expected_hash,
+                protocol_allowlist=protocol_allowlist,
+            )
+        )
+    return directory_path.absolute()
```
mHash1m, line 306: returning absolute path now as directory sources couldn't locate the directory

dffml/util/net.py
```diff
@@ -145,106 +144,20 @@ def sync_urlretrieve_and_validate(
...
             url,
             filename=str(target_path),
             protocol_allowlist=protocol_allowlist,
             reporthook=progress_reporthook,
         )
+        return path
     validate_file_hash(
         target_path, expected_sha384_hash=expected_sha384_hash,
     )
+    return str(target_path.absolute())
```
pdxjohnny, line 157: It looks like we previously had `cached_download()` giving a `pathlib.Path` object, let's maintain that

```suggestion
    return target_path.absolute()
```

Issue #1088, opened 2021-04-09
util : log : Need a  decorator which logs the Runtime of the function it wraps. 


programmer290399:
## Pain Point

Currently , to measure the performance or to make sure if a function/co-routine has not become significantly slower , there's no debug logging utility that can measure the runtime of a function. 

## Proposed Solution

Implement decorator to measure the time of the wrapped function/co-routine in `util/log.py`

programmer290399:
I am working on a PR , for this issue

PR #1101, opened 2021-04-14
util:log: Added log_time decorator

programmer290399:
* Fixes: #1088 
* Ready for review 
* Question regarding tests for this feature : is the current way of testing good enough or we need to use some other method to test it?
     - One other way could be to patch `logging.debug`


programmer290399:
Both the failing tests :  `Tests / test (., 3.7)` &  `Tests / test (., 3.8) ` have the same error `ERROR: Could not install packages due to an OSError: [Errno 28] No space left on device`

CHANGELOG.md
```diff
@@ -10,8 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Tutorial on how to load models dynamically
   https://intel.github.io/dffml/tutorials/models/load.html
 - Added download progressbar in `util/net.py`
+- Added `log_time` decorator for logging running time of functions/coroutines.
 - Usecase example notebook for "Moving between models"
 - Documentation and testing support for notebooks
+
```
pdxjohnny, line 16: Remove superfluous whitespace change

```suggestion
```

dffml/util/log.py
```diff
@@ -28,3 +32,38 @@ def get_download_logger(root_logger):
     duplicate_filter = DuplicateFilter()
     logger.addFilter(duplicate_filter)
     return logger
+
+
+def log_time(func):
+    """ 
+    Decorator that can take either coroutine or normal function 
+    
+    Found on Stack Overflow from Mikhail Gerasimov, only modified
```
pdxjohnny, line 41: If we include third party code in the main package it has to go through legal review again, which is not something that we can do at this point. This code isn't ideal anyway,  we really don't need to borrow from anywhere, we have all the pieces already.

We should be using a monotonic clock.

We should try to wrap any kind of function, coroutine, or generator, using the code we had here (or something similar or better)

https://github.com/intel/dffml/blob/b66b8bd298bfc1ba347afd5733bc36cc5ded1ab0/dffml/util/net.py#L172-L222

pdxjohnny (changes requested):

dffml/util/log.py
programmer290399: okay , I wasn't aware about that we need legal reviews for adding third party code, just saw something similar in `dffml/service/dev.py` in `remove_ansi_escape` method of `PinDeps` CMD , so I thought it would be fine to use code as it is from third party sources, if give proper credit, 

However I will re-write this little part and take care of this point in future.....

pdxjohnny: Ahhhhhh yes, thank you, that code should have been removed before release. Good catch

programmer290399: Updated :) , I think it should be fine now

pdxjohnny:
This PR doesn't include examples `>>> ` (you may need to look into `tests/test_docstrings.py` `wrap_` stuff). We also need to make sure it works for all function types, as per this reference: 
https://github.com/intel/dffml/blob/b66b8bd298bfc1ba347afd5733bc36cc5ded1ab0/dffml/util/net.py#L172-L222

programmer290399:
Ok , would look into this

pdxjohnny:
Ya good point, no reason to do generator functions, my bad.

Let's make sure we have the docstrings though

programmer290399:
waiting for #1132

PR #1115, opened 2021-05-14
dffml/model/scikit/tests : test_scikit_integration.py test_scikit.py …

deepakdinesh1123:
…: changed is to '=='


pdxjohnny:
Great, thanks!

PR #1116, opened 2021-05-16
docs: notebooks: Creating notebooks in dffml

mHash1m:


pdxjohnny (commented):
Looking great! Let's think about maybe some kind of generic testing for all notebooks, so we don't have to write tests for them individually.

mHash1m:
@pdxjohnny alright. Generic in what sense? Like running the same file? Or partly generic, running different asserts for different examples? I thought we need individual tests since example code can vary.

PR #1122, opened 2021-05-25
docs: arch: Mutable config properties

pdxjohnny:
- We started with NamedTuples, those had immutable properties
- Then we moved to dataclasses, we currently have properties and mutable
- Now we want to move to a dataclass or dataclass like approach where we default to having properties be immutable. We add a mutable keyword argument to field() which would signify that the config property is allowed to be modified without creating an entirely new instance of the config object.
  - Motivation here is that for model hyperparameter tuning we need to support in a standard way how properties of models can be tuned at runtime without instantiating an entirely new model class.
  - Another motivation which is father off is automl. When we go to do automl, whatever is “the” automl engine we implement will need to know which properties of model configs it can tweak in an automated fashion, and which things, for example the network architecture for neural networks, or even so basic as the directory property are not allowed be tweaked automatically.
- Implementation
  - Assume all properties are immutable unless explicitly marked as mutable
  - The class which uses the config must register a callback function to handle changes to config properties
https://github.com/intel/dffml/pull/1122/files#diff-aff838ca657169c089cf5d6e41383c0eaeba07312d4ca5355ac0bdf91d08fa00R80
  - Please comment on the PR if you have any questions comments concerns
    - There is for sure more stuff to iron out here
      - For example, what happens if the package version changes?
- This discussion stemmed from lack of clarity on model saving and loading. Hashim was unsure when implementing a tutorial what should happen if a model is instantiated in Python with certain hyperparameters but the model exists in a saved state on disk.
  - What currently happens (at least in the case of scikit) is that the saved state is load and the config properties are ignored essentially, because they are not used to instantiate the scikit model class.
  - There is another ADR within the referenced PR that talks about model saving and loading. We essentially decided that in memory properties should override saved properties provided that those properties are mutable. And this is where the mutable discucussion came from.
- We decided that these changes will need to take place as a part of Saahil’s project.
  - Hashim did some work on this https://github.com/intel/dffml/pull/1122#issuecomment-859773956
  - @sakshamarora1 can help coordinate on this across @mHash1m and @programmer290399

### DIFF
--- a/docs/arch/0003-Config-Property-Mutable-vs-Immutable.rst
+++ b/docs/arch/0003-Config-Property-Mutable-vs-Immutable.rst
@@ -1,0 +1,102 @@
+3. Config Property Mutable vs Immutable
+=======================================
+
+Date: 2021-08-04
+
+Status
+------
+
+Accepted
+
+Context
+-------
+
+This document and 0004-Model-Saving-and-Loading were drafted at the same
+time. See the 2021-05-25 Weekly Sync Meeting recording for more details.
+
+Through discussion about model config properties in relation to modification of
+hyperparameters we found that there are two kinds of config properties in
+general (as it relates to this document).
+
+- Immutable
+
+    - Properties whose values are fundamentally tied to the object the config is
+      used for. For the lifetime of the object.
+
+    - If these config properties change, a new instance of the object using the
+      config should be instantiated.
+
+    - Think of these like the concious or the soul of the object. You cannot
+      modify them, because then that would be a new object.
+
+- Mutable
+
+    - Properties whose values may could change over the lifetime of the
+      object, yet the object is still fundamentally the same thing.
+
+    - Think of these like the organs of the object. You could transplant an
+      organ from one person into another, but you didn't change who that person
+      is.
+
+Decision
+--------
+
+Similar to the way :py:class:`typing.NamedTuple` works, config properties will
+default to immutable. This way implementers of classes can assume they don't
+need to deal with state changes to config properties unless they've explicitly
+opted in.
+
+We will require that a callback function be registered with the config class.
+This way we ensure that modification of mutable values is handled somewhere.
+To combat class authors from declaring properties as mutable, without
+registering the callback. We'll raise an exception if the callback hasn't been
+registered and any property is set or retrieved, or the instance is exported.
+
+In the following example we have a hyperparameter, C. Out model wants to support
+changes to it at runtime, so we mark it as mutable.
+
+.. code-block:: python
+
+    @config
+    class MyModelConfig:
+        C: int = field("C hyperparameter", mutable=True)
+
+Since we've marked C as mutable, we must add a callback which deals with
+mutations to the config object.
+
+The callback we add accepts the key being modified within the config structure
+and the value it was set to. For demonstration purposes we will use the value of
+C in the model object. The callback's use of :py:func:`setattr` when the value
+of C is updated in the config structure results in the value of C in the model
+object also changing.
+
+.. code-block:: python
+
+    class MyModel(Model):
+        CONFIG = MyModelConfig
+        CONTEXT = ModelConfig
+
+        def __init__(self, config):
+            # Callback which will change self.C when self.config.C is changed
+            self.config.add_mutable_callback(lambda key, value: setattr(self, key, value))
+            # Use the value of C
+            self.C = self.config.C
+
+    model = MyModel(C=0)
+    assert model.C == 0
+    model.config.C = 42
+    assert model.C == 42
+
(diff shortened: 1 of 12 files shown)


docs/arch/0004-Model-Saving-and-Loading.rst
```diff
@@ -0,0 +1,101 @@
...
+Status
+------
+
+Draft
+
+Context
+-------
+
+See :doc:`0003-Config-Property-Mutable-vs-Immutable` for more context. These
+documents we started during the same Weekly Sync Meeting (2021-05-25).
```
pdxjohnny, line 15: https://youtu.be/y37CRHu8Oh4

docs/arch/0004-Model-Saving-and-Loading.rst
```diff
@@ -0,0 +1,101 @@
...
+.. todo::
+
+  - Need to figure out default values.
+
+    - Record that property should be default value?
+
+    - What happens if package version changes?
+
+      - Should we record what default value was and use old value in event of
+        change? Or should we use the new value?
```
pdxjohnny, line 73: Pinging @yashlamba, @sakshamarora1, @0dust, @mHash1m, @sk-ip, @programmer290399

mHash1m:
Tried out the [patch](https://gist.github.com/pdxjohnny/4d5ffddbd758b7729d544a2106e4efff) from our last [meeting](https://www.youtube.com/watch?v=aEQPbYpff80&t=2520s) on the [example](https://github.com/intel/dffml/pull/1122/files#diff-aff838ca657169c089cf5d6e41383c0eaeba07312d4ca5355ac0bdf91d08fa00R75-R86) in `Config Property Mutable vs Immutable`. 
It didn't work as-is (after implementing the getter) as the immutable setter gets called on default. I bypassed the exception to see what's happening and I get this:

```  
File "/home/hash1m/miniconda3/lib/python3.7/dataclasses.py", line 357, in wrapper
    result = user_function(self)
  File "<string>", line 3, in __repr__
  File "/home/hash1m/dffml/dffml/base.py", line 336, in getter
    return self._data[key]
KeyError: 'operations'
```
So I'm guessing we need to find a way to somehow not set these initial keys and avoid them so it doesn't blow up, right? 

Anyway, I tried setting the `setter` exclusively to `make_mutable` and the example seems to work fine (doesn't blow up and asserts pass).

So I have the following ambiguities:

- How does this work with the interface? Are we going to have the user give a flag in the config when instantiating the model ie. along with features and directory. In this case, what's the link between the parameter flag and our data class flag.
- I'm assuming we won't be explicitly setting each hyperparameter as mutable as we did in the example config(`C: int = field("C hyperparameter", mutable=True)`), so how'd that work?

ping @pdxjohnny

docs/arch/0003-Config-Property-Mutable-vs-Immutable.rst
```diff
@@ -0,0 +1,131 @@
...
+
+Consequences
+------------
+
+The :py:func:`field <dffml.base.field>` function will need to be modified to add
+a ``mutable`` keyword argument.
+
+.. todo::
+
+If a model is instantiated with config. Saved state in directory. Saved state
```
pdxjohnny, line 100: This stuff was copy pasted since these two docs started as one

pdxjohnny:
- Let’s have a way to have mutable=True be the default for every property
  - This can be implemented as a helper function which calls the config() decorator

docs/arch/0003-Config-Property-Mutable-vs-Immutable.rst
```diff
@@ -0,0 +1,101 @@
...
+The callback we add accepts the key being modified within the config structure
+and the value it was set to. For demonstration purposes we will use the value of
+C in the model object. The callback's use of :py:func:`setattr` when the value
+of C is updated in the config structure results in the value of C in the model
+object also changing.
+
+.. code-block:: python
+
+    class MyModel(Model):
+        CONFIG = MyModelConfig
```
pdxjohnny, line 76:
```suggestion
        CONFIG = MyModelConfig
        CONTEXT = ModelContext
```

mHash1m:
I pulled the branch and played with it.
Spent the last couple of hours trying to make it work on a preset config from a previous session :facepalm: 

Anyway, everything works perfectly fine while working in the same session.

Here's a little preview of how I verified mutable properties.
```
class ScikitConfig(ModelConfig, NamedTuple):
    C: int = field("C hyperparameter", mutable=True)
    tol: int = field("tol hyperparameter", mutable=True)
```
image

Instantiating with `C=0` and training blows up the model.
Then we set `model.config.C = 1` and it works :tada:  

This is great, thank you @pdxjohnny .

PR #1123, opened 2021-05-26
docs: conf: Fix .ipynb download button

mHash1m:
@pdxjohnny unfortunately the download .ipynb button failed on master docs.  Perhaps I didn't test it well enough.
I've managed to fix it.

Tested it with master and it directs to : https://intel.github.io/dffml/master/examples/notebooks/moving_between_models.ipynb
Tested it by making the if condition false and it directs to: https://intel.github.io/dffml/examples/notebooks/moving_between_models.ipynb

So it seems to be behaving right, now.

Thought about testing it on tag release too but there are too many changes to docs since then and the previous script wasn't running so not sure.
Let me know if I should figure that out.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -84,15 +84,15 @@ source_suffix = {
 }
 
 # Download button for ipython notebooks
-nb_version = ""
-if "." not in version:
-    nb_version = "master/"
 # This is processed by Jinja2 and inserted before each notebook
 nbsphinx_prolog = r"""
 {% set docname = env.doc2path(env.docname, base=None) %}
+{% if "." not in env.config.release %}
+    {% set nb_version = "master/" %}
+{% endif %}
 
 .. image:: ../../images/Download-.ipynb-button.svg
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -84,15 +84,15 @@
...
 # This is processed by Jinja2 and inserted before each notebook
 nbsphinx_prolog = r"""
 {% set docname = env.doc2path(env.docname, base=None) %}
+{% if "." not in env.config.release %}
+    {% set nb_version = "master/" %}
+{% endif %}
 
 .. image:: ../../images/Download-.ipynb-button.svg
-    :target: https://intel.github.io/dffml/{{nbversion}}{{docname[:-6]}}ipynb
+    :target: https://intel.github.io/dffml/{{ nb_version }}{{ docname[:-6] }}ipynb
```
mHash1m, line 95: in case you wonder why nb_version isn't initialized outside, it seems inside the jinja2 env if the variable isn't set, it will just ignore it and so it works.

PR #1127, opened 2021-06-10
examples: notebooks: Add use-case example for 'evaluating model performance' and script to create tests.

mHash1m:
# The Use-case Notebook
Changed name to `evaluating_model_performance` so in the future we can squash `tuning_models` into this one.
I think `moving_between_models` should stay as it is since it also has to do with getting predictions of models interchangeably so it cannot be highlighted if squashed.
On the other hand, comparing model performance and tuning models will go hand in hand.

# The `test_notebooks` script 

My concern is that the current assertions are dependant upon variables in the notebook. It won't blow up but it won't be helpful if someone goes ahead and names variables different.
I could go ahead and collect all the cell stdouts (which would require handling errors if the cell isn't a code cell) into one output of the notebook, but then again, the stdout could also vary in the sense that maybe someone doesn't bother printing out "Accuracy:" before the accuracy which would mean that there is no way to find it in the output. So the stdouts wouldn't really be a solution.

### A solution that comes to mind
We could maybe keep a template notebook, to maintain a standard and keep things uniform, in which we don't fill in the specifics but leave it for the user/contributor. like having the load() function empty to let the user fill in which model he wants and similarly with data and imports and everything. And at the end we have this Accuracy variable with our desired name or something like that and similarly with preds.

Other than that, the script seems to be working alright. Got the following log from some debugging statements showing which files the tests are created for and the directory they run in:

```
(base) hash1m@hash1m-ThinkPad-W520:~/dffml$ python3 -m unittest tests.test_notebooks
/home/hash1m/dffml
/home/hash1m/dffml/examples/notebooks/compare_model_performance.ipynb
/home/hash1m/dffml/examples/notebooks/moving_between_models.ipynb
/home/hash1m/dffml/tests/notebooks/data
./home/hash1m/dffml/tests/notebooks/data
.
Ran 2 tests in 12.222s

OK
```
ping @pdxjohnny


tests/test_notebooks.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+
+
+def find_notebooks():
+    return [path for path in nb_path.glob("*.ipynb")]
+
+
+def mk_notebook_test(path: pathlib.Path):
+    def testcase(self):
+        with chdir(pathlib.Path("tests/notebooks/data")):
```
pdxjohnny, line 26:
```suggestion
        with chdir(pathlib.Path(__file__).parent.joinpath("notebooks", "data")):
```

tests/test_notebooks.py
```diff
@@ -0,0 +1,53 @@
+import os
+import pathlib
+import unittest
+import logging
+
+from testbook import testbook
+
+from dffml.util.os import chdir
+
+nb_path = pathlib.Path(__file__).parent.parent / "examples/notebooks"
```
pdxjohnny, line 10:
```suggestion
nb_path = pathlib.Path(__file__).parent.parent / "examples" / "notebooks"
```

PR #1134, opened 2021-06-21
examples: notebooks: Create usecase example notebook 'Ensemble by stacking'

mHash1m:
- Using 2 models instead of 4.
- Accuracy for ensembling 2 models as suggested didn't turn out great.
- Removed the accuracies part to keep the notebook precise.
- Showing final predictions instead of ensemble accuracy comparison, requires `stacked_test_predictions` either way.

**Update**
- Fixed Accuracy
- Added Visualizations

Signed-off-by: mHash1m <hashimchaudry23@gmail.com>


pdxjohnny:
Pinging @sakshamarora1 for review

pdxjohnny:
This will need to be rebased for accuracy scorers and location

PR #1142, opened 2021-06-29
model: scikit: scorer: Add sklearn model accuracy scorer 

sk-ip:
Signed-off-by: Sudhanshu Sudhanshukumar5459@gmail.com

### DIFF
--- a/model/scikit/dffml_model_scikit/scikit_model_scorer.py
+++ b/model/scikit/dffml_model_scikit/scikit_model_scorer.py
@@ -1,0 +1,57 @@
+from dffml.base import config
+from dffml.util.entrypoint import entrypoint
+from dffml.source.source import SourcesContext
+from dffml.model import ModelContext, ModelNotTrained
+from dffml.accuracy import (
+    AccuracyScorer,
+    AccuracyContext,
+)
+
+
+class ScorerWillNotWork(Exception):
+    pass
+
+
+@config
+class SklearnModelAccuracyConfig:
+    pass
+
+
+class SklearnModelAccuracyContext(AccuracyContext):
+    """
+    Default scorer for Sklearn Models.
+    """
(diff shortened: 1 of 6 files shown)


model/scikit/dffml_model_scikit/scikit_model_scorer.py
```diff
@@ -0,0 +1,57 @@
...
+    Default scorer for Sklearn Models.
+    """
+
+    async def score(self, mctx: ModelContext, sctx: SourcesContext):
+        if not mctx._filepath.is_file():
+            raise ModelNotTrained("Train model before assessing for accuracy.")
+
+        if mctx.clf._estimator_type not in ("classifier", "regressor"):
+            raise ScorerWillNotWork(
+                "SklearnModelAccuracy will not work with Clustering Models"
```
pdxjohnny, line 31: Based on if statement let's make this:

```suggestion
                "SklearnModelAccuracy only works with Classifiers and Regressors"
```

Just in case there are more types in the future. Keeps the logic and the message the same

model/scikit/dffml_model_scikit/scikit_model_scorer.py
```diff
@@ -0,0 +1,57 @@
...
+    pass
+
+
+class SklearnModelAccuracyContext(AccuracyContext):
+    """
+    Default scorer for Sklearn Models.
+    """
+
+    async def score(self, mctx: ModelContext, sctx: SourcesContext):
+        if not mctx._filepath.is_file():
```
pdxjohnny, line 26: We should probably put this on the list of things that need to be exposed by models.

So far we already knew we would need to expose the equivalent of the `_estimator_type`. We should also be exposing the model's trained or not trained status. Tracking here: #1143

PR #1144, opened 2021-06-29
Accuracy Scorers

pdxjohnny:
Fixes: #779


.ci/deps.sh
```diff
@@ -56,6 +56,14 @@ if [[ "x${PLUGIN}" == "xsource/mysql" ]] || \
   docker pull mariadb:10
 fi
 
+if [[ "x${PLUGIN}" == "xmodel/tensorflow" ]]; then
```
pdxjohnny, line 59: Oops looks like we need to remove these

dffml/accuracy/accuracy.py
```diff
@@ -0,0 +1,60 @@
...
+        Returns
+        -------
+        float
+            The score value
+        """
+        raise NotImplementedError()
+
+
+@base_entry_point("dffml.accuracy", "accuracy")
+class AccuracyScorer(BaseDataFlowFacilitatorObject):
```
pdxjohnny, line 50: We need to make sure that we have consistent naming. We likely should remove the `Scorer` portion of the `AccuracyScorer` class name. Since it looks like most subclasses drop the `Scorer` suffix anyway.

Issue #1147, opened 2021-07-01
model: pytorch: pretrained: Add Python API support for defining the last layers when using the pre-trained Models


sakshamarora1:
We are assuming that the layers will be a nested dictionary of layers and its parameters which we convert into PyTorch layers using the `create_layer` function.

https://github.com/intel/dffml/blob/5edb093780f8ae9c76f34336f188482d9efc70d2/model/pytorch/dffml_model_pytorch/pytorch_pretrained.py#L63-L66

We need to implement it such that if `self.config.layers` is of type `nn.Sequential` or (`nn.ModuleList` or `nn.ModuleDict`, in  which case we should iterate through them and create a sequential layer) - [reference to pytorch docs](https://pytorch.org/docs/stable/nn.html#containers), which we finally use to replace the last layer of the pre-trained model

https://github.com/intel/dffml/blob/5edb093780f8ae9c76f34336f188482d9efc70d2/model/pytorch/dffml_model_pytorch/pytorch_pretrained.py#L68-L83

PR #1148, opened 2021-07-02
model: pytorch: pretrained: Add support for additional layers Python API

mHash1m:
fixes: #1147 

* This PR is a dependency for the usecase example notebook I'm currently working on, ie. "Transferlearning".

* I ended up setting `layers: Any`in the `PyTorchPreTrainedModelConfig` and also skipping `convert_value()` in `base.py` if field type is `Any` since it tries to create an instance of the type. 

* Seems to be working as I handle the formatting of the parameter in `pytorch_pretrained` itself, but what could be the implications of this? 

* It appears we don't have `Any` type fields, even if we add in future and want to by-pass the checks in `base.py`, we can simply provide multiple arguments to `isinstance()` to check against along with `dict`, to allow the `convert_value()` process.

* Alternatively, I could also create a `.yaml` inside the notebook (for the time being)  if this approach doesn't seem right.

* Using `Pillow==8.2.0` to temporarily fix the pytorch errors. Seems like there's a pending fix python-pillow/Pillow#5572 so should be okay soon

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Operations for file (de)compression
 - Usecase example notebook for "Evaluating Model Performance"
 - Tests for all notebooks auto created and run via ``test_notebooks.py``
+- Support for additional layers in pytorch pretrained models via Python API
 ### Changed
 - Calls to hashlib now go through helper functions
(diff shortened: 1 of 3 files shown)


pdxjohnny:
Pinging @sakshamarora1 and or @yashlamba for review

pdxjohnny:
- [ ] Add releated PR number to commit message trailer

PR #1149, opened 2021-07-06
operations: data: Add data cleanup operations

sk-ip:
Signed-off-by: Sudhanshu Sudhanshukumar5459@gmail.com


pdxjohnny:
- [ ] Let's re-run package creation from skel/
- [ ] Include an example .rst to test out, let's use at least two operations as
      a part of this

dffml/operation/source.py
```diff
@@ -0,0 +1,92 @@
...
+
+@config
+class SourceOperationConfig:
+    source: BaseSource
+
+
+@op(
+    inputs={
+        "features": Definition(name="features", primitive="List[str]"),
+        "predict_features": Definition(
```
pdxjohnny, line 17: We don't seem to use `predict_features` anywhere, what's the point in having it?

dffml/source/dfpreprocess.py
```diff
@@ -0,0 +1,114 @@
+import pathlib
+from typing import AsyncIterator, List
+
+from dffml.base import config, field
```
pdxjohnny, line 4: Need relative imports

pdxjohnny (commented):
We're only using one cleanup operation here. We should make more tutorials or use more cleanup ops in this one. I like this tutorial as is. I think making more might the best plan.

We don't nessicarily need to train a model even, just show how to run the ops. Let's make sure we show how to use at least 2 more ops, ideally together in the same flow

dffml/source/dfpreprocess.py
sk-ip: ok

sk-ip:
@pdxjohnny , I forgot to push one more examples which I had written this was somehow got untracked.

dffml/operation/source.py
sk-ip: We are actually using this when we do not need to process some of the features in the dataset, like the prediction value or the classification value.

sk-ip:
[kc_house_data.csv](https://github.com/intel/dffml/files/7040979/kc_house_data.csv)

sk-ip:
[mushrooms.csv](https://github.com/intel/dffml/files/7040983/mushrooms.csv)

sk-ip:
[kc_house_data.csv](https://github.com/intel/dffml/files/7046671/kc_house_data.csv)

PR #1151, opened 2021-07-06
examples: notebooks: Create use-case example notebook for 'Transfer Learning'

mHash1m:
**Progress :**
- [x] Squashed unnecessary part.
- [x] Fixed timeout during training.
- [x] Fixed macos error by installing and enabling `ipywidgets`.
- [x] Enhance code descriptions.
- [x] Fix features.
- [x] Remove the finetuning approach
- [x] Add links to documentation for cached_download, source.dir, and model entry points
- [x] move `ipywidgets` to dev `extra_require`
- [x] ignore warnings in jupyter notebook
- [x] set logging to `INFO` and stream to `stdout` to avoid the red background.
- [x] Hide the logging cell from docs
- [x] Remove the personal user handle from paths in JSON so it doesn't show on docs


mHash1m:
@pdxjohnny @sakshamarora1 can we please get this one reviewed first as I haven't had any feedback on this one yet.

sakshamarora1 (changes requested):
Awesome work!
Since I can't comment on the notebook changes, I'll point out here a few things/doubts that might need to be a bit more elaborate:

- A bit more explanation on the `cached_download_unpack_archive` i.e. why are we giving the arguments we are giving to it.
- Instead of doing `Model.load` to use the entrypoint-loading, I think it would make more sense to import it from `dffml_model_pytorch`
- What does the `Accuracy2` signify here?
- I think we should show the train output here, to make it more insightful.
- Also, the feature extraction example doesn't have the end layers replaced to accommodate the rockpaperscissors use-case. Similarly, in the fine-tuning example, the `out_features` should be 3 and not 17. Both the cases are almost the same with just the `trainable` bool parameter being true and false in them. So maybe we should show either one, just like here - https://intel.github.io/dffml/master/examples/flower17/flower17_pytorch.html 

I love the `display_image_predictions` function here! Great stuff. @pdxjohnny this is the kind of function I wanted to add as an operation for using in CLI too, but didn't got around to it. I guess it works best in the Python API example.

mHash1m:
> Instead of doing `Model.load` to use the entrypoint-loading, I think it would make more sense to import it from `dffml_model_pytorch`

@pdxjohnny wanted dynamic loading to be used in these tutorials. 

> What does the `Accuracy2` signify here?

Made quite a lot of changes, doesn't make sense afterwards. Thanks for the catch.

> I think we should show the train output here, to make it more insightful.

I don't get any outputs on train at all, other than warnings about my CUDA and my local paths(say "hashim/dffml/" etc) so I cleared the train outputs.
Wasn't sure if this is dffml and jupyter interaction or if outputs are to be expected at all.

> Also, the feature extraction example doesn't have the end layers replaced to accommodate the rockpaperscissors use-case.
 
Yeah, I wasn't sure whats the default on it and interaction between pytorch and dffml, so I asked you about that one on gitter.

> Similarly, in the fine-tuning example, the `out_features` should be 3 and not 17. 

Thanks for the catch, I was using the flower dataset so forgot to adjust that one shifting to rps. The other features are wrong as well.

> Both the cases are almost the same with just the `trainable` bool parameter being true and false in them. So maybe we should 
show either one, just like here - https://intel.github.io/dffml/master/examples/flower17/flower17_pytorch.html

That does make sense @pdxjohnny final thoughts on this?

Thanks @sakshamarora1 for the amazing review! I'll update these changes asap.

pdxjohnny:
Can we move `ipywidgets` to the `dev` `extras_require`?

sakshamarora1:
> I don't get any outputs on train at all, other than warnings about my CUDA and my local paths(say "hashim/dffml/" etc) so I cleared the train outputs.
> Wasn't sure if this is dffml and jupyter interaction or if outputs are to be expected at all.

Importing the logging module and setting it to INFO will work as far as I think. Try playing with that.

mHash1m:
- [x] move `ipywidgets` to dev `extra_require`
- [x] ignore warnings in jupyter notebook
- [x] set logging to `INFO` and stream to `stdout` to avoid the red background.
- [x] Hide the logging cell from docs
- [x] Remove the personal user handle from paths in JSON so it doesn't show on docs

@pdxjohnny @sakshamarora1 the logs seem good now.

Issue #662, opened 2020-05-20
model: Rename directory property to location


pdxjohnny:
# Support Archive Storage for Models

The goal of this project is the modify Models and DataFlows to support saving
and loading from archives (such as `.zip` and `.tar` for example).

Currently all DFFML models have a `directory` property, which is where they
store the contents of the model's state. It does not currently hold a copy of
the model's config.

We'll want to modify the Model's context entry and exit code to have it
pack/unpack it's config and its saved state directory from/into the archive at
`location`.

## Skills

- Python

## Difficulty

Beginner/Intermediate

## Related Readings

- https://intel.github.io/dffml/master/docs/contributing/gsoc/2021/

## Getting Started

- Read the contributing guidelines
  - https://intel.github.io/dffml/master/contributing/index.html
- Go through the quickstart
  - https://intel.github.io/dffml/master/quickstart/model.html
- Go trough the model tutorials
  - https://intel.github.io/dffml/master/tutorials/models/
- Go through the model plugins
  - https://intel.github.io/dffml/master/plugins/dffml_model.html
- Start programming to understand better what will need to happen. The following
  is a rough guide meant to give you some activities that will help you better
  understand what needs to be done so you can write a more complete proposal.
  - Rename the `directory` property to `location` for all models.
  - On `__aenter__()` of `dffml.model.model.Model`, if `location` has an
    extension, such as zip or tar, extract it to a temporary directory.
  - Write a method, `get_directory()`, which returns the location of the
    temporary directory property on the `Model` class.
  - On `__aexit__()` , pack the model state directory into the `location`, along
    with it's config.
  - Understand where all the code for models are that will potentially need to
    be changed (`dffml/model` and `model/`)
  - Make this work with test cases for the SLR model.

## Potential Mentors

- [John Andersen](https://github.com/pdxjohnny)
- [Yash Lamba](https://github.com/yashlamba)
- [Saksham Arora](https://github.com/sakshamarora1)

## Tracking and Discussion

This project is related to the following issues. Please discuss and ask
questions in the issue comments. Please also ping mentors on
[Gitter](https://gitter.im/dffml/community) when you post on the following
issues so that they are sure to see that you've commented.

- https://github.com/intel/dffml/issues/662

sudharsana-kjl:
@pdxjohnny I'd like to work on this issue

pdxjohnny:
[...]

PR #1155, opened 2021-07-10
WIP : model: Rename directory property to location

programmer290399:
* Second step in Fixing: #662


dffml/model/model.py
```diff
@@ -2,12 +2,16 @@
 # Copyright (c) 2019 Intel Corporation
 """
 Model subclasses are responsible for training themselves on records, making
-predictions about the value of a feature in the record, and assessing thei
+predictions about the value of a feature in the record, and assessing their
```
pdxjohnny, line 5: Good job putting changes like this in separate commits

dffml/model/model.py
```diff
@@ -80,28 +88,87 @@ def __init__(self, config):
...
+
+        dataflow = self._create_dataflow(
+            operation,
+            {
+                "input_directory_path": input_path,
+                "output_file_path": output_path,
+            },
+        )
+
+        async for _, _ in MemoryOrchestrator.run(dataflow):
```
pdxjohnny, line 143: Let's use the `high_level:run` function

dffml/model/model.py
```diff
@@ -80,28 +88,87 @@ def __init__(self, config):
...
+        if self.location_is_file:
+            config_path = self.config.location / "config.json"
+            config_path.write_text(json.dumps(self.config))
+            output_location = self.config.file_location
+            await self._run_operation(
+                "archive", self.temp_dir, output_location
+            )
+            shutil.rmtree(self.temp_dir)
+
+    async def _run_operation(self, action, input_path, output_path):
```
pdxjohnny, line 123: - [ ] Let's do the rename in a separate PR than the implementation of the dataflow stuff
  - [ ] Looking at this, It seems like the best way to make progress is to wait to do the rename of directory to location until we have the dataflow stuff implemented. Since we can implement using `self.directory` and then do the rename later.

dffml/model/model.py
programmer290399: I actually used that before but it was not able to import it in `model.py` file probably because of some circular dependency..... I was getting the error cannot import name run from high_level

dffml/model/model.py
programmer290399: So shall make this PR only for implementation of the dataflow stuff ? and yes the rename can be done after that.....

dffml/model/model.py
programmer290399: A lot of errors like this : 
```
ImportError: Failed to import test module: test_slr
  File "/usr/lib/python3.7/unittest/loader.py", line 154, in loadTestsFromName
    module = __import__(module_name)
  File "/home/saahil/rps/dffml/tests/model/test_slr.py", line 4, in <module>
    from dffml import SLRModel, run_consoletest, AsyncTestCase
  File "/home/saahil/rps/dffml/dffml/__init__.py", line 69, in <module>
    for import_name, module in modules(root, package_name, skip=skip):
  File "/home/saahil/rps/dffml/dffml/__init__.py", line 30, in modules
    yield import_name, importlib.import_module(import_name)
  File "/usr/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "/home/saahil/rps/dffml/dffml/high_level.py", line 11, in <module>
    from .model.model import Model
  File "/home/saahil/rps/dffml/dffml/model/__init__.py", line 18, in <module>
    from .model import Model, ModelContext, SimpleModel, ModelNotTrained
  File "/home/saahil/rps/dffml/dffml/model/model.py", line 26, in <module>
    from ..high_level import run
ImportError: cannot import name 'run' from 'dffml.high_level' (/home/saahil/rps/dffml/dffml/high_level.py)
```

programmer290399: I think it might be due to some circular dependency loop being formed .... 🤔

dffml/model/model.py
programmer290399: ok sounds perfect to me :)

pdxjohnny:
**This is in progress I'm editing it right now, will remove this comment when I'm done**


Here's my thought process here.

- From the changes we have it's clear that we need to implement the dataflow stuff, then do the rename in a separate PR, or else this will be way to hard to debug / review.
  - We can implement by looking at the `config.directory` property first, and then rename everything later.
  - When I look at the `__aenter__()` code where we make the dataflow and do the extract, and set the location to the extracted directory, this is what comes to mind.
    - If we modify the `config.location` property, then we loose the original value, if a user were to copy the config and instantiate another model class using it, they would get essentially a different model, one that points to the tempdir rather than the original archive path.
      - Therefore, the location / directory should really be a variable tied to the model object instance (`self`), rather than the model object's config.
      - We should probably go through the existing models and modify them 



In light of this, I see this as the likely steps for this stage of your project. Each top level bullet point being a different PR.

- Implement dataflows
  - 2 functions, they take an archive format as their input.
    - One for makeing a dataflow that does extraction of a given archive format

programmer290399:
* Only one thing is failing in CI and i.e. [`Tests / lint (lines, 3.7, 12.x)`](https://github.com/intel/dffml/runs/3042268357?check_suite_focus=true)
```bash
$ python3 ./scripts/check_literalincludes.py
'docs/tutorials/dataflows/chatbot.rst' might need updating as line numbers of 'examples/dataflow/chatbot/operations.py' may have changed
'docs/tutorials/models/docs.rst' might need updating as line numbers of 'dffml/skel/model/REPLACE_IMPORT_PACKAGE_NAME/myslr.py' may have changed
This script checks to see if any .py files changed that are referenced in .rst files using specific line numbers. This script is failing because those rst files were not also modified. This script is not smart enough to tell if those lines should be modified, or that you modifed them correctly. Just as an extra sanity check.

You'll notice that the literalinclude blocks might end up with the text in them being off because the :lines: wasn't changed. That's what this script is hoping to catch.
```
* I found out that this is not an issue because the line numbers  haven't really changed and the master docs and docs from this branch look identical for the two files listed above ....

pdxjohnny:
https://asciinema.org/a/425823

- Commit message formatting
- Remove wine_quality.csv

Also found two more instances

```patch
diff --git a/service/http/docs/api.rst b/service/http/docs/api.rst
index fd808d876..f350d0fc2 100644
--- a/service/http/docs/api.rst
+++ b/service/http/docs/api.rst
@@ -239,7 +239,7 @@ The following is an example request body to configure a model. The URL this
     "model": {
       "plugin": null,
       "config": {
-        "directory": {
+        "location": {
           "plugin": [
             "/home/user/modeldirs/mymodel"
           ],
diff --git a/service/http/docs/python.rst b/service/http/docs/python.rst
index 797e6ca24..ef2a61bb6 100644
--- a/service/http/docs/python.rst
+++ b/service/http/docs/python.rst
@@ -27,7 +27,7 @@ This is an example of how you can use the web API from Python.
       "model": {
         "plugin": null,
         "config": {
-          "directory": {
+          "location": {
             "plugin": [
               "/home/user/modeldirs/mymodel"
             ],
```

Checked using:

```console
pdxjohnny@rza dffml $ git grep directory=
dffml/service/dev.py:                http.server.SimpleHTTPRequestHandler, directory=str(pages_path)
dffml/util/testing/consoletest/commands.py:        handler_class = functools.partial(handler_class, directory=directory)
docs/arch/0002-Object-Loading-and-Instantiation-in-Examples.rst:        directory="model",
docs/examples/dataflows.rst:          directory=key \
docs/installation.rst:                "https://github.com/intel/dffml/archive/master.zip#egg=dffml-feature-git&subdirectory=feature/git"
docs/installation.rst:                "https://github.com/intel/dffml/archive/master.zip#egg=dffml-feature-git&subdirectory=feature/git"
feature/git/dffml_feature_git/feature/operations.py:            URL=repo.URL, directory=repo.directory, commit=commit
pdxjohnny@rza dffml $ git grep \"directory\"
dffml/operation/archive.py:DIRECTORY = Definition(name="directory", primitive="str")
dffml/util/testing/consoletest/commands.py:            hasattr(other, "directory") and self.directory == other.directory
dffml/util/testing/consoletest/commands.py:            hasattr(other, "directory") and self.directory == other.directory
dffml/util/testing/consoletest/commands.py:            hasattr(other, "directory") and self.directory == other.directory
docs/examples/dataflows.rst:          '[{"pypi_package_contents": "directory"}]="dffml.mapping.create".inputs.value' \
examples/shouldi/shouldi/deploy/df/shouldi.json:                "directory": [
examples/shouldi/shouldi/deploy/df/shouldi.json:                        "pypi_package_contents": "directory"
examples/shouldi/shouldi/deploy/df/shouldi.json:                "directory": "cleanup_pypi_package.inputs.directory"
examples/shouldi/shouldi/deploy/df/shouldi.json:                "directory": "run_bandit.inputs.pkg"
examples/shouldi/shouldi/python/pypi.py:    outputs={"directory": run_bandit.op.inputs["pkg"]},
examples/shouldi/shouldi/python/pypi.py:            return {"directory": package_src_dir}
examples/shouldi/shouldi/python/pypi.py:    inputs={"directory": run_bandit.op.inputs["pkg"]}, stage=Stage.CLEANUP,
examples/shouldi/shouldi/use.py:                            "directory": str(path),
examples/shouldi/tests/test_pypi.py:            no_files = os.listdir(results["directory"])
examples/shouldi/tests/test_pypi.py:            shutil.rmtree(results["directory"])
feature/git/dffml_feature_git/feature/operations.py:    return {"repo": {"URL": URL, "directory": directory}}
operations/deploy/tests/test_operations.py:        return {"repo": {"URL": URL, "directory": directory}}
pdxjohnny@rza dffml $ git grep \\-directory
dffml/util/testing/consoletest/commands.py:        if "--directory" in self.cmd:
dffml/util/testing/consoletest/commands.py:            directory = self.cmd[self.cmd.index("--directory") + 1]
docs/contributing/docs.rst:    $ python3 -m http.server --directory pages/ 7000
docs/contributing/testing.rst:    $ python3 -m http.server --directory htmlcov/ 8080
```

I'm putting with changes into master now...

pdxjohnny:
Merged as 5c1d437eac2b2e48a20866664faef607a42fe2d9

PR #1174, opened 2021-07-27
model : Add Support for Archive Storage of Models

programmer290399:
* Fixes: #662 
* Have updated all the models, this branch is ready for final review and merge. 
* No tests are failing except commit lint and I have updated the enhancement issue as it shouldn't.
* TODO:
    - [x] Add Tutorial on Archive Support 
    - [x] Add tests for newly added `dffml.df.archive` 
    - [x] Update SLR model 
    - [x] Update Scikit model 
    - [x] Update Scratch model 
    - [x] Update Daal4py model 
    - [x] Update Xgboost model 
    - [x] Update Auto Sklearn model 
    - [x] Update Spacy model 
    - [x] Update Vowpal Wabbit model 
    - [x] Update Pytorch model 
    - [x] Update Tensorflow model 
    - [x] Update Tensorflow hub  model
    - [x]  Remove any debugging related code


dffml/model/model.py
```diff
@@ -84,6 +106,208 @@ def __call__(self) -> ModelContext:
...
+            self.location = temp_dir
+
+    async def __aexit__(self, exc_type, exc_value, traceback):
+        if self.config.location.is_file():
+            config_path = self.location / "config.json"
+            config_path.write_text(json.dumps(export(self.config)))
+            await self._run_operation(
+                "archive", self.temp_dir, self.config.location
+            )
+            shutil.rmtree(self.temp_dir)
```
pdxjohnny, line 124: Let's make sure that we remove the tempdir if we created it, even if location is no longer a file

dffml/model/model.py
```diff
@@ -84,6 +106,208 @@ def __call__(self) -> ModelContext:
...
+            )
+            self.location = temp_dir
+
+    async def __aexit__(self, exc_type, exc_value, traceback):
+        if self.config.location.is_file():
+            config_path = self.location / "config.json"
+            config_path.write_text(json.dumps(export(self.config)))
+            await self._run_operation(
+                "archive", self.temp_dir, self.config.location
+            )
```
pdxjohnny, line 123: Let's make sure that if this is the first time we are using a model, then we create an archive of the saved state

dffml/model/model.py
```diff
@@ -84,6 +106,208 @@ def __call__(self) -> ModelContext:
         self._make_config_location()
         return self.CONTEXT(self)
 
+    async def __aenter__(self):
+        if self.config.location.is_file():
```
pdxjohnny, line 110: - Let's add two more properties to the model config, Let's check for them in the same way we check of location (`hasattr()`), don't assume that a model config has a location property, don't assume that it has `location_save` or `location_load`. These will be of type DataFlow. Don't go add these to the each model yet. Just implement for SLR model to test first. We'll add to other model configs in next PR.

- Let's take the dataflows you're creating, and put the helper function to create them defined in `dffml/dataflow/archive.py`. When we `run()` the dataflow we can pass the `input_path` and the `output_path`. Let's make sure that we always pass these to the dataflow as some standard definition name. Maybe, `model_location` and `model_tempdir` or something, the flow will need to be updated to grab from these locations for the dataflows your function creates.

dffml/model/model.py
programmer290399: okay, would push this in a future commit

dffml/model/model.py
```diff
@@ -84,6 +106,208 @@ def __call__(self) -> ModelContext:
         self._make_config_location()
         return self.CONTEXT(self)
 
+    async def __aenter__(self):
```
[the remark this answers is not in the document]
pdxjohnny: Let's follow what we wrote in the Saving and Loading models ADR, and if it doesn't make sense please comment / review in that PR so we write how it should work there, then implement here.
https://github.com/intel/dffml/pull/1186/files#diff-100d7fa4cb67a1eb6648d6a81054f58a20cc8e7333d91d519faed60487240b81R87-R101

programmer290399:
Blank Diagram (1)
@pdxjohnny @sakshamarora1  please give your inputs.

dffml/model/model.py
programmer290399: Mentioned related flow [here](https://github.com/intel/dffml/pull/1174#issuecomment-893176136)

dffml/model/model.py
programmer290399: Done

dffml/model/model.py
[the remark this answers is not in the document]
programmer290399: Ok , but the thing is how shall we load the non-conflicting ones in because I guess our @config doesn't allow adding new attributes .....

dffml/model/model.py
```diff
@@ -84,6 +106,208 @@ def __call__(self) -> ModelContext:
...
+        if self.config.location.is_file():
+            temp_dir = self._get_directory()
+            await self._run_operation(
+                "extract", self.config.location, temp_dir
+            )
+            self.location = temp_dir
```
pdxjohnny:
```suggestion
            self.location = temp_dir
        return self
```

dffml/model/model.py
```diff
@@ -87,6 +98,85 @@ def __call__(self) -> ModelContext:
...
+            )
+            # Load values from config if it exists
+            config_path = self.temp_dir / "config.json"
+            if config_path.exists():
+                with open(config_path) as config_handle:
+                    loaded_config = json.load(config_handle)
+                    for prop, value in loaded_config.items():
+                        # TODO: Need to figure out how loading of
+                        # properties should be done.
+                        pass
```
pdxjohnny, line 128: Let's leave this as a TODO for now, but within this loop body, instead of let's set anything that is of the correct type. Let's `self.logger.warning` and log if there is a differnce.

dffml/model/model.py
programmer290399: Done, thanks 🙂 , that was really helpful ......

dffml/model/model.py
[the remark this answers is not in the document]
programmer290399: Relevant Discussion  [Weekly Sync 2021-08-10 @ 1:11:43](https://youtu.be/okQvV-R_cBA?t=4303)

TLDR: The saving and loading stuff would be handled in a generic manner later on , for now we'd be moving forward with [this](https://github.com/intel/dffml/pull/1174#discussion_r686189809)

dffml/model/model.py
programmer290399: Done 👍🏽

dffml/model/model.py
programmer290399: The flow was implemented and pushed, which handles the first time use case as well using mimetypes...

dffml/model/model.py
```diff
@@ -89,10 +89,10 @@ def __init__(self, config):
...
             # to treat "~" as the the home location rather than a literal
             location = location.expanduser().resolve()
-            # TODO Change all model configs to make them support mutable
-            # location config properties
-            with self.config.no_enforce_immutable():
-                self.config.location = location
+        # TODO Change all model configs to make them support mutable
+        # location config properties
+        with self.config.no_enforce_immutable():
+            self.config.location = location
```
pdxjohnny, line 95: Why has this been moved in 4 spaces?

programmer290399: I moved it out 4 spaces, I am not sure why it was within the ` if isinstance(location, pathlib.Path):` body, but it was creating problems in some test cases where this block was not executed, leaving the `self.config.location` attribute  a string whereas we expect it to be a `pathlib.Path`

pdxjohnny:
Where is this at right now? We're closing on GSoC deadline and want to make sure this gets completed

dffml/df/archive.py
```diff
@@ -0,0 +1,214 @@
...
+    gz_compress,
+    gz_decompress,
+    bz2_compress,
+    bz2_decompress,
+    xz_compress,
+    xz_decompress,
+)
+
+
+def get_key_substr(string: str, dict: dict, return_="value") -> Any:
```
pdxjohnny, line 25:
```suggestion
def get_key_substr(string: str, dict: dict, return_value: bool = True) -> Any:
```

programmer290399: Have made this change is a local commit would push it soon .... :)

dffml/model/model.py
programmer290399: Have added this code in a local commit, would push this soon :)

dffml/df/archive.py
programmer290399: Done 👍🏽

dffml/model/model.py
programmer290399: Done 👍🏽

programmer290399:
@pdxjohnny 
# This should be merged **BEFORE** #1199 

#### Because this branch has the older version of #1199 merged in it,  doing otherwise would remove the latest changes from #1199 , not merging latest changes from #1199 into this branch because the tests take around 6 hours to complete since swportal tests were included.

pdxjohnny:
Merged as https://github.com/intel/dffml/commit/38390a45ea5a2aacd1dc2398ac2acbcf9b8c7a53

PR #1175, opened 2021-07-27
model: scikit: Add support and usecase examples for MultiOutput models

mHash1m:
- Remove last commit(d7ef74d07368450d33f0923afb77fef87d4d118c)? i.e native accuracy scorer multi-output support. 
    - Currently, we are trying to get mean of accuracies for the multiple labels which yields different output than the scikit scorer with same name. 
    - I have 3 approaches, the other two commented in the file, and all of them give different scores. Probably the math is off with the other 2.
- test (., 3.7), test (., 3.8) and examples/dataflows.rst take forever.
    - [x] fixed by removing the numpy import.
- [x] Add exceptions of scorers for multioutput support in docs.
- [x] Make high-level accuracy() take predict features.
- [x] Add docs if changes are finalized.
- [x] READY FOR REVIEW
- [ ] READY TO MERGE


model/scikit/dffml_model_scikit/scikit_base.py
```diff
@@ -113,22 +153,29 @@ async def predict(
...
-                self.parent.config.predict.dtype(self.clf.predict(predict)[0])
-                if self.parent.config.predict.dtype is not str
-                else self.clf.predict(predict)[0],
-                self.confidence,
-            )
+            target = self.predictions
+            if self.is_multi:
+                record.predicted(
+                    target, self.clf.predict(to_predict)[0], self.confidence,
+                )
```
pdxjohnny, line 168: Let's revert the change in record.prediected for array support, move the for loop to these lines

service/http/dffml_service_http/routes.py
```diff
@@ -911,7 +911,9 @@ async def model_predict(self, request, mctx):
     async def scorer_accuracy(self, request, mctx, actx):
         sctx_label_list = await request.json()
         sources = await self.get_source_contexts(request, sctx_label_list)
-        return web.json_response({"accuracy": await actx.score(mctx, sources)})
+        return web.json_response(
+            {"accuracy": await actx.score(mctx, mctx.config.predict, sources)}
```
mHash1m, line 915: alternative for this?
Adding this fixed test_cli but test_routes failing now

pdxjohnny:
We decided that the accuracy scorers should take the features to score as their
last argument(s) to the `score()` method.

In this way, if a scorer wanted to work on multiple features, the method
signiture could be easily modified to be
```
    async def score(
        self,
        mctx: ModelContext,
        sources: SourcesContext,
        *args: Feature,
    ):
```


```diff
diff --git a/dffml/accuracy/mse.py b/dffml/accuracy/mse.py
index aeb00fb82..ed28d3d24 100644
--- a/dffml/accuracy/mse.py
+++ b/dffml/accuracy/mse.py
@@ -21,18 +21,17 @@ class MeanSquaredErrorAccuracyContext(AccuracyContext):
     Mean Squared Error
     """
 
-    async def score(self, mctx: ModelContext, sources: SourcesContext):
-        if len([mctx.parent.config.predict]) != 1:
-            raise InvalidNumberOfFeaturesError(
-                f"{self.__class__.__qualname__} can only assess accuracy of one feature. features: {features}"
-            )
+    async def score(
+        self,
+        mctx: ModelContext,
+        sources: SourcesContext,
+        feature: Feature,
+    ):
         y = []
         y_predict = []
         async for record in mctx.predict(sources):
-            y.append(record.feature(mctx.parent.config.predict.name))
-            y_predict.append(
-                record.prediction(mctx.parent.config.predict.name).value
-            )
+            y.append(record.feature(feature.name))
+            y_predict.append(record.prediction(feature.name).value)
         accuracy = sum(
             list(map(lambda x, y: abs(x - y) ** 2, y, y_predict))
         ) / len(y)
```

pdxjohnny:
These changes are required to get scorers working with HTTP service.

We need to add a **TODO** to the HTTP service refactor to accpect the features
to score from the request body.

```diff
diff --git a/service/http/dffml_service_http/routes.py b/service/http/dffml_service_http/routes.py
index e11d82b3f..5aaa3a596 100644
--- a/service/http/dffml_service_http/routes.py
+++ b/service/http/dffml_service_http/routes.py
@@ -912,7 +912,11 @@ class Routes(BaseMultiCommContext):
         sctx_label_list = await request.json()
         sources = await self.get_source_contexts(request, sctx_label_list)
         return web.json_response(
-            {"accuracy": await actx.score(mctx, mctx.config.predict, sources)}
+            {
+                "accuracy": await actx.score(
+                    mctx, mctx.parent.config.predict, sources
+                )
+            }
         )
 
     async def api_js(self, request):
diff --git a/service/http/dffml_service_http/util/testing.py b/service/http/dffml_service_http/util/testing.py
index 4fa9e1cf8..4cc36cd0f 100644
--- a/service/http/dffml_service_http/util/testing.py
+++ b/service/http/dffml_service_http/util/testing.py
@@ -1,4 +1,5 @@
 import asyncio
+import tempfile
 import contextlib
 from http import HTTPStatus
 from unittest.mock import patch
@@ -63,7 +64,9 @@ class FakeScorerConfig:
 
 
 class FakeScorerContext(AccuracyContext):
-    async def score(self, mctx: ModelContext, sources: SourcesContext):
+    async def score(
+        self, mctx: ModelContext, feature: Feature, sources: SourcesContext
+    ):
         accuracy: int = 0
         async for record in sources.records():
             accuracy += int(record.key)
@@ -178,11 +181,20 @@ class TestRoutesRunning:
 
     @contextlib.asynccontextmanager
     async def _add_fake_model(self):
-        async with FakeModel(BaseConfig()) as model:
-            self.model = self.cli.app["models"][self.mlabel] = model
-            async with model() as mctx:
-                self.mctx = self.cli.app["model_contexts"][self.mlabel] = mctx
-                yield
+        with tempfile.TemporaryDirectory() as tempdir:
+            async with FakeModel(
+                FakeModelConfig(
+                    location=tempdir,
+                    features=Features(Feature("by_ten")),
+                    predict=Feature("by_ten"),
+                )
+            ) as model:
+                self.model = self.cli.app["models"][self.mlabel] = model
+                async with model() as mctx:
+                    self.mctx = self.cli.app["model_contexts"][
+                        self.mlabel
+                    ] = mctx
+                    yield
 
     @contextlib.asynccontextmanager
     async def _add_fake_scorer(self):
```

mHash1m:
> ```diff
> diff --git a/dffml/accuracy/mse.py b/dffml/accuracy/mse.py
> index aeb00fb82..ed28d3d24 100644
> --- a/dffml/accuracy/mse.py
> +++ b/dffml/accuracy/mse.py
> @@ -21,18 +21,17 @@ class MeanSquaredErrorAccuracyContext(AccuracyContext):
>      Mean Squared Error
>      """
>  
> -    async def score(self, mctx: ModelContext, sources: SourcesContext):
> -        if len([mctx.parent.config.predict]) != 1:
> -            raise InvalidNumberOfFeaturesError(
> -                f"{self.__class__.__qualname__} can only assess accuracy of one feature. features: {features}"
> -            )
> ```

I'm assuming here that these changes in `mse` were just to show how a scorer for multioutput would differ and not something to be implemented into `mse` as mse doesn't support multioutput and removing this logic and sending `Features` to `mse` results in the following error:

image

So I'm going to leave the logic as is, please let me know if we're not on the same page here @pdxjohnny .

dffml/util/python.py
```diff
@@ -175,3 +175,27 @@ def within_method(obj: object, method_name: str, max_depth: int = -1) -> bool:
         ):
             return True
     return False
+
+
+def no_inplace_append(list_a: list, arg_b: Union[str, list]) -> list:
```
pdxjohnny, line 180: Isn't this the same as `list_a + [arg_b]`?

```python
>>> a = [1]
>>> id(a)
140385771636928
>>> b = a + [2]
>>> id(b)
140385748642624
```

dffml/cli/ml.py
```diff
@@ -64,6 +66,13 @@ async def run(self):
...
         if inspect.isclass(self.scorer):
             self.scorer = self.scorer.withconfig(self.extra_config)
+        if hasattr(self.model.config, "predict"):
+            return await accuracy(
+                self.model,
+                self.scorer,
+                self.model.config.predict,
+                self.sources,
+            )
         return await accuracy(self.model, self.scorer, self.sources)
```
pdxjohnny, line 76: We need to modify this command to take the features to assess accuracy from the command line (via `AccuracyCMDConfig`)

dffml/high_level/ml.py
```diff
@@ -140,6 +159,17 @@ async def accuracy(
             # through something like pydantic. See issue #36
             raise TypeError(f"{accuracy_scorer} is not an AccuracyScorer")
         # Run accuracy method
+        # if accuracy scorer expects predict feature(s), pass as argument else don't
+        if (
+            accuracy_scorer.__class__.__qualname__
+            not in NO_PREDICT_FEATURE_SCORERS
+        ):
+            if "features" in locals():
```
pdxjohnny, line 167: Instead of checking if features is in locals, it is more Pythonic to set features to None, and then check `if features is not None`

dffml/high_level/ml.py
```diff
@@ -140,6 +159,17 @@ async def accuracy(
             # through something like pydantic. See issue #36
             raise TypeError(f"{accuracy_scorer} is not an AccuracyScorer")
         # Run accuracy method
+        # if accuracy scorer expects predict feature(s), pass as argument else don't
+        if (
+            accuracy_scorer.__class__.__qualname__
+            not in NO_PREDICT_FEATURE_SCORERS
```
pdxjohnny, line 165: The main module really shouldn't have to know anything about plugins, let's remove `NO_PREDICT_FEATURE_SCORERS`.

dffml/high_level/ml.py
```diff
@@ -125,6 +137,13 @@ async def accuracy(
     Accuracy: 0.0
     """
     async with contextlib.AsyncExitStack() as astack:
+        # if predict feature(s) were passed into the model, extract and set args for sources
+        if isinstance(args[0], (Feature, Features)):
+            features = args[0]
+            args = args[1:]
+        elif isinstance(args[-1], (Feature, Features)):
+            features = args[0]
+            args = args[:-1]
```
pdxjohnny, line 146: We have overloaded the possible types of the variable length arguments too far with this implementation. Let us instead modify the function signature of `accuracy()` to accept either a `Feature`, or `Features` as the third argument.

```python
async def accuracy(
    model,
    accuracy_scorer: Union[AccuracyScorer, AccuracyContext],
    features: Union[Feature, Features],
    *args: Union[BaseSource, Record, Dict[str, Any], Feature, Features],
) -> float:
```

dffml/high_level/ml.py
pdxjohnny: But let's remove this, we should always be passing features via expansion.

```python
return await actx.score(mctx, sctx, *features)
```

dffml/high_level/ml.py
```diff
@@ -140,6 +159,17 @@ async def accuracy(
...
+            accuracy_scorer.__class__.__qualname__
+            not in NO_PREDICT_FEATURE_SCORERS
+        ):
+            if "features" in locals():
+                return float(await actx.score(mctx, sctx, features))
+            else:
+                raise PredictFeatureExpected(
+                    f"{accuracy_scorer} expects predict feature(s) as an argument. Please refer to docs"
+                )
         return float(await actx.score(mctx, sctx))
```
pdxjohnny, line 173:
```suggestion
        return await actx.score(mctx, sctx, *features)
```

dffml/high_level/ml.py
pdxjohnny, line 146: We have overloaded the possible types of the variable length arguments too far with this implementation. The existing logic is not very clean.

Let us instead modify the function signature of `accuracy()` to accept either a `Feature`, or `Features` as the third argument.

```python
async def accuracy(
    model,
    accuracy_scorer: Union[AccuracyScorer, AccuracyContext],
    features: Union[Feature, Features],
    *args: Union[BaseSource, Record, Dict[str, Any], Feature, Features],
) -> float:
```

```suggestion
    # TODO Use this to ensure that we're always passing features before records
    # We can remove it eventually once we know we've updated everywhere
    # appropriately
    if not isinstance(features, Feature):
        raise TypeError(f"features was {type(features)}: {features!r}. Should have been Feature or Features")
    if isinstance(features, Feature):
        features = Features(features)
    async with contextlib.AsyncExitStack() as astack:
```

dffml/high_level/ml.py
pdxjohnny: Why was this logic added?

dffml/accuracy/accuracy.py
```diff
@@ -25,7 +26,7 @@ def __init__(self, parent: "Accuracy") -> None:
 
     @abc.abstractmethod
     async def score(
-        self, mctx: ModelContext, sources: SourcesContext,
+        self, mctx: ModelContext, sources: SourcesContext, feature: Feature
```
pdxjohnny, line 29:
```suggestion
        self, mctx: ModelContext, sources: SourcesContext, *args: Feature
```

dffml/accuracy/clf.py
```diff
@@ -21,16 +21,18 @@ class ClassificationAccuracyContext(AccuracyContext):
...
-    async def score(self, mctx: ModelContext, sources: SourcesContext):
-        if len([mctx.parent.config.predict]) != 1:
+    async def score(
+        self, mctx: ModelContext, sources: SourcesContext, feature: Feature,
+    ):
+        if len([feature]) != 1 or isinstance(feature, Features):
             raise InvalidNumberOfFeaturesError(
-                f"{self.__class__.__qualname__} can only assess accuracy of one feature. features: {features}"
+                f"{self.__class__.__qualname__} can only assess accuracy of one feature. Refer to docs to find a scorer that supports Multi-Output"
             )
```
pdxjohnny, line 30: We really don't need this check logic if we call with list argument expansion (`*features`) from `high_level.ml:accuracy()`.

dffml/accuracy/mse.py
```diff
@@ -21,18 +22,18 @@ class MeanSquaredErrorAccuracyContext(AccuracyContext):
...
-    async def score(self, mctx: ModelContext, sources: SourcesContext):
-        if len([mctx.parent.config.predict]) != 1:
+    async def score(
+        self, mctx: ModelContext, sources: SourcesContext, feature: Feature,
+    ):
+        if len([feature]) != 1 or isinstance(feature, Features):
             raise InvalidNumberOfFeaturesError(
-                f"{self.__class__.__qualname__} can only assess accuracy of one feature. features: {features}"
+                f"{self.__class__.__qualname__} can only assess accuracy of one feature. Refer to docs to find a scorer that supports Multi-Output"
             )
```
pdxjohnny, line 31: Same as clf with no need to do this check

dffml/cli/ml.py
pdxjohnny: Once we do that we should only have one possible call to `accuracy()` here.

```python
        return await accuracy(self.model, self.scorer, self.features, self.sources)
```

dffml/high_level/ml.py
mHash1m: To accommodate spacy ner but it seems like it doesn't have python api.

dffml/util/python.py
mHash1m: True lol :facepalm:

dffml/cli/ml.py
mHash1m: I suppose we dont want the cli to take predict twice, I think we should use the '-model-predict' arg for features but I'm not sure where to look in the arg parsers for that, can you point me?

pdxjohnny: Hmmm, we really should do it twice, we had a lot of issues when we inferred the directory to save things in. We need to make sure that everything is explicate rather than implicit

pdxjohnny:
Nice!!!

Great work @mHash1m! 💯 🥇 🌟

PR #1177, opened 2021-07-28
ci: run: Fix for infinite run of commit msg lint on master

programmer290399:
* I am not sure what the branch name is for master inside the GitHub actions env 
* It  may or may not be `master` ,  I cannot verify it directly as I cannot push to master and check, 
* If it is different than `master` then one small change would be required [here](https://github.com/intel/dffml/pull/1177/files#diff-7b436cb5923ba7f66ccf2e140b342330400dcb1d6776d03444564dc34c5379f1R194) , just replace master with appropriate branch name in the env

### DIFF
--- a/.ci/run.sh
+++ b/.ci/run.sh
@@ -189,7 +189,7 @@ function run_style() {
 }
 
 function run_commit(){
-  BRANCH="$(git rev-parse --abbrev-ref HEAD)"
+  BRANCH="$(echo $GITHUB_REF | cut -d'/' -f 3)"
   echo "On Branch: ${BRANCH}"
   if [[ "$BRANCH" != "master" ]]; then
     dffml service dev lint commits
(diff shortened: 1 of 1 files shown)


programmer290399:
ping @pdxjohnny

.ci/run.sh
```diff
@@ -189,7 +189,11 @@ function run_style() {
 }
 
 function run_commit(){
-  dffml service dev lint commits
+  BRANCH="$(echo $GITHUB_REF | cut -d'/' -f 3)"
+  echo "On Branch: ${BRANCH}"
+  if [[ "$BRANCH" != "master" ]]; then
```
pdxjohnny, line 194:
```suggestion
  if [[ "x${BRANCH}" != "xmaster" ]]; then
```

programmer290399: I didn't do that because I am pretty sure the GitHub Actions probably uses bash and as it is  a POSIX-compliant shell this (afaik) makes the added x comparison redundant 🤔

programmer290399: However if there's any other reason to add this then please let me know as well.

PR #1186, opened 2021-08-04
docs: arch: Model Saving and Loading

pdxjohnny:
Co-authored-by: mHash1m <hashimchaudry23@gmail.com>
Co-authored-by: Saahil Ali <programmer290399@gmail.com>
Co-authored-by: Sudhanshu kumar <sudhanshukumar5459@gmail.com>

### DIFF
--- a/docs/arch/0004-Model-Saving-and-Loading.rst
+++ b/docs/arch/0004-Model-Saving-and-Loading.rst
@@ -1,0 +1,101 @@
+4. Model Saving and Loading
+===========================
+
+Date: 2021-05-25
+
+Status
+------
+
+Draft
+
+Context
+-------
+
+See :doc:`0003-Config-Property-Mutable-vs-Immutable` for more context. These
+documents we started during the same Weekly Sync Meeting (2021-05-25).
+
+We currently have inconsistencies in the models are saved and loaded. Due to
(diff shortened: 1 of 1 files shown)


mHash1m:
@pdxjohnny Is this PR supposed to be completed to properly use mutable config stuff even in a single session with #1122 ? 

It seems like I missed something. I had set `mutable = True`and I thought the parameters I was changing to test didn't have much effect on the accuracy as I was testing on a small dummy dataset then. So I tried to blow up the model while instantiating and then setting the parameter to something normal. 

While that worked, it doesn't work once the model is trained(the values still get passed to the model) as we are saving the model and there's currently no way to mutate a saved model, even if it's trained again.

Could this be simply fixed by finding what method was called inside __aenter__ (I'm not sure how to do this, couldn't find anything googling a few minutes) and not loading the saved model if its `train()` and overwriting the saved model?
Temporarily tried it directly in train to test it out and it seems to do trick.
```
    async def train(self, sources: Sources):
        if self._filepath.is_file():
            os.remove(str(self._filepath))
            config = self.parent.config._asdict()
            del config["location"]
            del config["predict"]
            del config["features"]
            self.clf = self.parent.SCIKIT_MODEL(**config)
```

I'm even able to load a model cross-session and do with parameters as I please.

What are the repercussions for this?

PR #1189, opened 2021-08-11
docs: arch: Object Saving and Loading

pdxjohnny:


docs/arch/0006-Object-Saving-and-Loading-with-DataFlows.rst
```diff
@@ -0,0 +1,42 @@
+6. Object Saving and Loading with DataFlows
```
pdxjohnny, line 1: Discussion here: https://github.com/intel/dffml/discussions/1216

PR #1195, opened 2021-08-17
ci: run: consoletests: Fail on error

pdxjohnny:
- ci: run: consoletests: Fail on error
- examples: tutorials: models: slr: run: Add MSE scorer


pdxjohnny:
Merged

Issue #1198, opened 2021-08-20
operation: archive: Directory structure not preserved while creating a zip/tar archive of a directory


programmer290399:
## Describe the bug

While creating an archive the directory structure is not preserved. 

## Steps To Reproduce

Try archiving any directory or use following example.

<details><summary>Example by @pdxjohnny </summary>
<p>

```py
import os
import sys
import pathlib
import tempfile
import itertools
from typing import Union, Callable

from dffml.operation.archive import (
    make_zip_archive,
    extract_zip_archive,
    make_tar_archive,
    extract_tar_archive,
)
from dffml.operation.compression import (
    gz_compress,
    gz_decompress,
    bz2_compress,
    bz2_decompress,
    xz_compress,
    xz_decompress,
)

from dffml.df.types import DataFlow, Input, Operation
from dffml.df.base import OperationImplementation, op
from dffml.noasync import run

# Must include output operation for results (from run()) to contain data!
from dffml.operation.output import GetSingle


def make_dataflow(
    *args: Union[Operation, OperationImplementation, Callable], **kwargs
) -> DataFlow:
    return DataFlow(
        # Add an output operation. Results will be empty without one.
        GetSingle,
        # Add the operations to the dataflow.
        *args,
        seed=[
            # Ensure the output operation returns a single value for each
            # operation executed that has an output. If operations will be run
            # multiple times and produce multiple outputs which you want to
            # capture, you'll want to add GetMulti rather than GetSingle.
            Input(
                value=list(
                    itertools.chain(
                        *[
                            [
                                definition.name
                                for definition in operation.op.outputs.values()
                            ]
                            for operation in args
                        ]
                    )
                ),
                definition=GetSingle.op.inputs["spec"],
            )
        ]
        + [
            # Add each input from kwargs.
            Input(
                value=value,
                # Look up the definition which should be used this value. The
                # name of the keyword argument will be used to look up the
                # definition. This is just a quick and dirty approach to this,
                # obviously if there are two inputs with the same operation
                # local name, only one of their definitions will be chosen.
                definition=dict(
                    itertools.chain(
                        *[
                            [
                                (input_name, definition)
                                for input_name, definition in operation.op.inputs.items()
                            ]
                            for operation in args
                        ]
                    )
                )[key],
            )
            for key, value in kwargs.items()
        ],
    )


@op
def debug_print_tempdir(input_directory_path: str):
    """
    Debug function to print contents of tempdir
    """
[...]

PR #1199, opened 2021-08-20
operation: archive: patched to preserve dir structure

programmer290399:
* Fixes: #1198 
* Partially FIxes: #662

### DIFF
--- a/dffml/operation/archive.py
+++ b/dffml/operation/archive.py
@@ -1,7 +1,8 @@
-from os import name
+import os
 import tarfile
 import zipfile
 import pathlib
+
 
 from ..df.base import op
 from ..df.types import Definition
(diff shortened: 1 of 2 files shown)


programmer290399:
Ping @pdxjohnny for review.

dffml/operation/archive.py
```diff
@@ -41,8 +42,16 @@ async def make_zip_archive(
         Path to the output zip file
     """
     with zipfile.ZipFile(output_file_path, "w") as zip:
-        for file in pathlib.Path(input_directory_path).rglob("*"):
-            zip.write(file, file.name)
+        # TODO: Make a clean implementation of this with pathlib
+        for root, _, files in os.walk(input_directory_path):
```
pdxjohnny, line 46: Let's look to https://github.com/python/cpython/blob/7ef0673de48958bc3a75af5f152564bd2dffa8dd/Lib/zipfile.py#L2366-L2424 as a reference

programmer290399: Shall I make an enhancement issue for this ? 
As it would take time to verify the changes and I need to hit the deadline. 
It works fine currently , just the implementation is not ideal I'd say, 
going for a better implementation would require extra time and testing, 
so I will try to do it as soon as possible after the deadline ......
does that sound good to you ?

programmer290399: Ok, looking into it , would push a similar solution asap .....

programmer290399: by the way can we create an enhancement issue for this and come back to it later ?

programmer290399: Done, let me know if any other changes are required .....

programmer290399:
# This should be merged **AFTER** #1174 

#### Because it has the older version of this branch merged in it, so doing otherwise would remove the latest changes from this branch, not merging this branch into it because the tests take around 6 hours to complete since swportal tests were included.

PR #1000, opened 2021-02-18
model: pytorch Update the call to numpy array creation

iamapickle:
My First Pull Request, kindly check and inform if I have committed any mistakes.

Issue #987, opened 2021-02-12
util: testing: consoletest: 


pdxjohnny:
In `dffml/util/testing/consoletest/`, we have the ability to read `.rst` files and run the console tests in them (see the README for more info).

We can parse out code blocks with the code in `parser.py`.

The goal of this issue is to put a light wrapper on the CLI functionality to accept a document without `code-block` directives, for example, the body of this commit message: bc681e28bba2be39308fa3e3a9292c62fcecc5f3.

You probably want to add a new argument to `cli.py`, `--code-block-content` which is a boolean that tells us if the input we are getting is not an `.rst` file, but rather, the content of a `code-block`.

You
Looking at the code in `util`, you might be able to modify `cli.py` to have `nodes = parse_nodes(...)`

Vk24-hash:
Hey @pdxjohnny I am new to open source contribution, can I give this one a try?

pdxjohnny:
@Vk24-hash please do! Make sure to check out the contributing docs for some guidance.

Vk24-hash:
Thanks, I read the issue and what i got from it was that we have to make
add control statements to a new argument passed in cli.py which would call
util.code_block_to_dict() if it's true and nodes = parse_nodes(...) if it's
false . I am confused as if I have to make a new function or edit an
existing one, if I am supposed to edit then how to know which one I have to
edit? Please guide me.

pdxjohnny:
Start by reading the code, and running the example in the README.md in the directory where the files we're talking about are located. Then try editing the function where `parse_nodes()` is being used.

```console
$ git grep -n parse_nodes -- dffml/util/testing/consoletest/cli.py
dffml/util/testing/consoletest/cli.py:7:from .parser import parse_nodes
dffml/util/testing/consoletest/cli.py:47:    for node in parse_nodes(args.infile.read()):
```

Vk24-hash:
Thanks, I'll check it out.

PR #1208, opened 2021-08-25
Updated pytorch_base.py

krishsethi19:
Updated the call to numpy array creation #1000 
Added ability to parse input as if it was code-block's conent #987


model/pytorch/dffml_model_pytorch/pytorch_base.py
```diff
@@ -126,10 +126,11 @@ async def dataset_generator(self, sources: Sources):
             )
         if not y_cols:
             raise ValueError("No records to train on")
-
+        np.warnings.filterwarnings('ignore', category=np.VisibleDeprecationWarning)
```
pdxjohnny, line 129: We can't just turn off warnings, we have to fix what it is warning us about

```suggestion
        np.warnings.filterwarnings('ignore', category=np.VisibleDeprecationWarning)
```

dffml/util/testing/consoletest/cli.py
```diff
@@ -43,17 +44,24 @@ async def main(argv: List[str]) -> None:
...
-    for node in parse_nodes(args.infile.read()):
-        if not node.options.get("test", False):
-            continue
-        if node.directive == "code-block":
-            nodes.append(
-                code_block_to_dict(node.content, node.options, node=node.node)
-            )
-        elif node.directive == "literalinclude":
+    if args.code_block_content :
+
```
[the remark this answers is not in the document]
pdxjohnny: We need an associated test case.

dffml/util/testing/consoletest/cli.py
```diff
@@ -43,17 +44,24 @@ async def main(argv: List[str]) -> None:
...
+                nodes.append(
+                    literalinclude_to_dict(node.content, node.options, node.node)
+                )
+    else :
+        for node in parse_nodes(args.infile.read()):
             nodes.append(
-                literalinclude_to_dict(node.content, node.options, node.node)
-            )
+                    code_block_to_dict(node.content, node.options, node=node.node)
+                )
```
pdxjohnny, line 64: The issue says that we need to introduce functionality to parse input as if it was the contents of a code block.

```python
            nodes.append(
                code_block_to_dict(args.infile.read(), node.options, node=node.node)
            )
```

pdxjohnny (changes requested):

PR #1061, opened 2021-04-01
docs: arch: 2nd and 3rd party plugins

pdxjohnny:
Here is the start on the CI changes for the main package: https://gist.github.com/pdxjohnny/ae1430367040298b3a895abc4d2863ee

PR #1273, opened 2021-11-07
util: testing: manifest: shim: Initial commit

pdxjohnny:
We're involved in several CI/CD projects at the moment. This concept of a manifest seems important (see [2ndparty](https://github.com/intel/dffml/pull/1061) dependency declaration section for background on how this applies to DFFML, it applies to other projects as well.

If we (talking to non-DFFML CI/CD community people here) use a shim layer it sets us up for forward compatibility. We can all work to define manifest formats that work well for our projects. By using the shim we ensure that they share enough properties that we set ourselves up for an easier eventual merging for manifest formats into one common CI/CD manifest format.

From the DFFML side, a manifest can easily be represented as a DataFlow. The manifest is a more user friendly form of a dataflow essentially, something more tailored to a given use case which describes which operations should be preformed, in what order, and what data types are involved. The shim layer allows projects to parse the manifest however they want. It allows them to easily switch what their next phase is at any time. This next phase might be their existing workflow, or it might be a dataflow enabled workflow. If the next phase parser outputs a dataflow when given a manifest we'll ensure auditability and achievability of the subsequent phase processing that will be taken.

The shim layer effectively operates as a trampoline (think UEFI shim, kexec, etc.) bridging the gap between the initial phase of manifest parsing and the next phase (which in DFFML's case is looking like metric collection, ml job execution, and executing our own CI both in the upcoming polyrepo setup as well as enabling local testing).

### DIFF
--- a/dffml/util/testing/manifest/shim.py
+++ b/dffml/util/testing/manifest/shim.py
@@ -498,6 +498,13 @@ DEFAULT_INPUT_ACTIONS = {
 }
 
 
+class InputActionNotFound(Exception):
+    """
+    Input actions are used to read in manifest. If one is not found then the
+    manifest cannot be read.
+    """
+
+
 class ParserNotFound(Exception):
     """
     Document format/version/action combination not found. It was not registered
@@ -718,6 +725,12 @@ def shim(
         ).values()
     }
     # Determine how to get the manifest
+    if args.input_action not in input_actions:
+        raise InputActionNotFound(
+            "Input action is used to read in manifest"
+            f" {args.input_action!r} not found in loaded input actions:"
+            f" {input_actions!r}"
+        )
     input_action = input_actions[args.input_action]
     # Get the manifest and any validation data that might be associated with it
     contents = input_action(args)
--- a/dffml/util/testing/manifest/shim.py
+++ b/dffml/util/testing/manifest/shim.py
@@ -6,15 +6,13 @@ Test Procedure Specification (TPS) Report Manifest Shim
 Validate and parse a Test Procedure Specification (TPS) Report manifest. Execute
 something for the next stage of parsing.
 
-This file is used as a shim to bridge the gap between the parsing for the
-TPS manifest format and the next action to taken after parsing. This file allows
-for registration of next phase parsers via environment variables.
-
-The purpose of this script is to preform the initial validation and parsing of
-the TPS manifest. It's responsibility is to then call the appropriate next phase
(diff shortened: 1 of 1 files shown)


dffml/util/testing/manifest/shim.py
```diff
@@ -0,0 +1,704 @@
...
+
+
+def popen_write_to_stdin(
+    cmd: List[str], write: bytes, **kwargs
+) -> subprocess.Popen:
+    """
+    Call :py:func:`subprocess.Popen`
+    """
+    read_end, write_end = os.pipe()
+    proc = subprocess.Popen(cmd, stdin=read_end, **kwargs)
```
pdxjohnny, line 138: doctest and `Popen.communicate` were not playing nice together, so when with pipe for now

pdxjohnny:
https://www.agwa.name/blog/post/ssh_signatures

dffml/util/testing/manifest/shim.py
```diff
@@ -0,0 +1,1459 @@
...
+        document_version,
+        schema_url,
+        schema_validation_action,
+        schema_validation_target,
+    ) = parse_schema_url(manifest["$schema"])
+    # Validate schema
+    if not args.no_schema_check:
+        raise NotImplementedError(
+            f"schema validation not implemented {(schema_url, schema_validation_action, schema_validation_target)!r}"
+        )
```
pdxjohnny, line 1321: schema validation goes here

probably want to add `only_schema_check` here

dffml/util/testing/manifest/shim.py
```diff
@@ -0,0 +1,1459 @@
...
+version of the format. Then the rest should be the extension which is
+associated with the format which we can use to validate the contents
+of the document, such as JSON schema.
+
+Optionally the schema URL may include a hash after which a key value pair may
+specify a validation action and target respectively.
+
+.. code-block:: yaml
+
+    $schema: "https://example.com/my.document.format.0.0.0.schema.json#sha256=dfd781dca25694dd808630ff4f27f290ba394143bdcae02911bcafa4a13b8319"
```
pdxjohnny, line 375: Could treat `.` as namespace

dffml/util/testing/manifest/shim.py
pdxjohnny: **TODO** validate `@context` for https://github.com/w3c/vc-data-model/issues/953

pdxjohnny: https://multiformats.io/

Issue #1264, opened 2021-10-24
ci: master: Multiple tests failing 


programmer290399:
All the failing tests have been listed below:

- [x] **test (model/scikit, 3.8)**

- [x] **test (model/scikit, 3.7)**

- [ ] **test (examples/shouldi, 3.8)**

- [ ] **test (examples/shouldi, 3.7)**

- [ ] **test (., 3.8)**

- [ ] **tutorials (3.7, docs/examples/dataflows.rst)**

- [ ] **tutorials (3.7, docs/examples/icecream_sales.rst)**

- [ ] **tutorials (3.7, docs/tutorials/dataflows/nlp.rst)**

- [ ] **test (., 3.7)**

programmer290399:
UPDATE: `tests.test_scikit.TestRANSACRegressorModel` fails randomly sometimes due to negative accuracy. More information up [here](https://stackoverflow.com/questions/56621252/negative-accuracy-score-in-regression-models-with-scikit-learn).

Need to look into it as well...

programmer290399:
# Latest list: 
> To be merged with the tasks above.

lint (style, 3.7, 12.x)

test (configloader/image, 3.8)

test (configloader/image, 3.7)

lint (docs, 3.7, 12.x)

test (model/scikit, 3.7)

test (examples/shouldi, 3.8)

test (examples/shouldi, 3.7)

test (model/scikit, 3.8)

test (., 3.8)

tutorials (3.7, docs/examples/dataflows.rst)

tutorials (3.7, docs/cli.rst)

tutorials (3.7, docs/examples/icecream_sales.rst)

tutorials (3.7, docs/tutorials/dataflows/nlp.rst)

test (., 3.7)

test (model/autosklearn, 3.8)

test (model/autosklearn, 3.8)

PR #1226, opened 2021-09-18
high_level: ml: Renamed `accuracy` to `score`

programmer290399:
* Fixes: #1181 
* Ready for review.

PR #1292, opened 2022-01-23
model: scikit: tests: Fixed failing scikit tests 

programmer290399:
* Related: #1264 
* PR is Ready for Review, Ping:@pdxjohnny   
* This bug was introduced in PR #1226, where I removed a seemingly redundant code block [here](https://github.com/intel/dffml/pull/1226/commits/ff096ae4ddec127815b6459e4ed5eedc92291f91#diff-be04a13bbc3837205b593fd019f43a1573590f980b28210e5d914879901eecc5L26-L28), after discussing with @sk-ip on gitter as shown below:
scikit_test_bug_origin
* Well he might not have realized that it wasn't there by mistake, it was supposed to override the baseclass implementation of `test_02_predict` as it was irrelevant for this test.
* I have fixed the test and also have added a comment describing why it has been left blank.

### DIFF
--- a/model/scikit/tests/test_scikit.py
+++ b/model/scikit/tests/test_scikit.py
@@ -183,6 +183,7 @@ class TestScikitModel:
                     else:
                         prediction = record.prediction(target).value
                     if self.MODEL_TYPE == "CLASSIFICATION":
+                        print("prediction:", prediction)
                         self.assertIn(prediction, [2, 4])
                     elif self.MODEL_TYPE == "REGRESSION":
                         correct = FEATURE_DATA_REGRESSION[int(record.key)][3]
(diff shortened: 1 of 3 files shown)


programmer290399:
https://github.com/sphinx-doc/sphinx/issues/4961

programmer290399:
Failing tests on this PR:

* `Tests / lint (changelog, 3.7, 12.x) (pull_request)` is failing because there's nothing to add there.
* `Tests / lint (docs, 3.7, 12.x) (pull_request)` has suddenly started failing, though it has nothing to do with this PR, and the issue is with sphinx I guess: https://github.com/sphinx-doc/sphinx/issues/4961, as the doc build succeeds locally on my machine with the old dependencies, so it looks like we need to pin some version of it...

* `Tests / test (., 3.8) (pull_request)`  is failing due to some issue in Ipython stuff in `dffml/util/python.py` and that is not related to this PR, rather it seems like a dependency version issue...

pdxjohnny:
Fantastic work Saahil!

PR #1330, opened 2022-03-13
docs: arch: DataFlow as Class

pdxjohnny:


docs/arch/0004-DataFlow-as-Class.rst
```diff
@@ -0,0 +1,201 @@
...
+
+  - Map arguments and keyword arguments to inputs and definitions
+
+  - Allow for adding a predefined set of inputs for a given method
+
+  - Inputs may also come from seed, auto started operations, or input networks
+
+  - DataFlow with output operations which will be used to query outputs from
+    input network context of the class's dataflow.
+
```
pdxjohnny, line 65:
```suggestion

  - Similar to context def with ability to pass clas instance as value for input
```

Issue #1336, opened 2022-03-16
docs: broken links in README.md and CONTRIBUTING.md


agriyakhetarpal:
1. The link mentioned in [https://github.com/intel/dffml/blob/f76f2ce94c3972634053377b00e7c16530f7f0a4/README.md#contributing](url) is broken and leads to an error 404 page.
2. The link mentioned in [https://github.com/intel/dffml/blob/f76f2ce94c3972634053377b00e7c16530f7f0a4/CONTRIBUTING.md](url) is also broken, just like the above one.

agriyakhetarpal:
I'd like to fix this. I'll submit a PR in a moment.

Issue #984, opened 2021-02-10
cleanup: Rename master branch to main


pdxjohnny:
There are a lot of references to `master` through out our codebase. `git grep` will be your friend here.

version0chiro:
I would like to work on this issue, if only the renaming of the 'master' references to 'main' is required or I need to do some checkout?

pdxjohnny:
@version0chiro Try to start re-naming things, push a PR, and we'll see what happens

rajpratyush:
@pdxjohnny I have tried renaming in my local branch but there was no pr generated . Can you assist me

rajpratyush:
@pdxjohnny  doesn't it requires admin access??

RupeshRavindraDharme:
Can you assign this issue to me
I would like to try

pdxjohnny:
This is shaping up to be fairly complex. We’re going to split out things into the second party plugin #1061 repo structure. When we do that we should probably do this at the same time. I’ve removed the good first issue tag

PR #1337, opened 2022-03-16
Fixed broken links

agriyakhetarpal:
This resolves issue #1336  wherein links referencing to the documentation and contributing guidelines were broken.


pdxjohnny:
This actually due to #984 which still isn't quite done yet, we need to fix the docs build to fix this, these URLs are correct, the deployed version of the docs is incorrect. But thank you

PR #1357, opened 2022-04-10
Fixed file hierarchy issue noticed in iris.py

TirelessClock:
Hey @programmer290399 @pdxjohnny,
I pulled the project onto my local device (py3 version 3.8.9) and tried running file iris.py, but got this error:

ImportError: attempted relative import with no known parent package

For the first import itself. I tried several things, including using Gitpod interface (py3 version 3.8.12), but the issue persisted. 
I identified the issue to be related to the file hierarchy, that is, trying to import a file in a higher directory.
I've solved the issue for this instance of the problem using the sys and path libraries, but this problem's present in several other parts of the project in lower level functional directories, I really hope we could look into this.

Please inform if there's an error in my personal implementation of the project. 
Thank you


programmer290399:
can you share how you executed the code and got this error?
Also, share from which directory you ran the code ...

TirelessClock:
Hey @programmer290399 ,
I was trying to understand the iris.py file, and in the process I ran it on my local device. In the process, I got the error (ImportError: attempted relative import with no known parent package). 
I ran it from the iris.py file itself, that is: dffml/ dffml/ source/ dataset/ iris.py

I hope that's what you were asking.

programmer290399:
Yes, and that is not how it works, as the iris.py is a module within dffml and not a standalone script, it uses relative imports accordingly, so to see how to use the iris dataset take a look at [this example](https://dffml.github.io/dffml-pre-image-removal/master/api/source/dataset/iris.html#module-dffml.source.dataset.iris).

If you are interested in knowing why this error comes up and how relative imports work in python see [this](https://napuzba.com/a/import-error-relative-no-parent).

I hope this helps...other than this I have one more thing, for solving any issue, please consider making an issue in the issue tracker, discuss your approach to solving the issue with the dffml team and then raise a PR to solve the issue, this will save you time and put your hard work to good use.

As you can see that this PR has no use for anyone as it is not an issue at all, I hope you'd understand, the dffml team is more than happy to help you out on gitter.

So I will kindly request you to close this PR. 🙂

TirelessClock:
Ahh, I understand now @programmer290399 .
See, the error came up, I tried a lot of different things and I couldn't understand what was happening, I assumed it was a problem. Sorry about that, will be sure to make an Issue next time.

programmer290399:
It's perfectly fine, no issues.

Issue #1343, opened 2022-03-19
docs: missing screenshot of grading rubric


Saransh-cpp:
The screenshot of the grading rubric seems to be missing in https://intel.github.io/dffml/master/contributing/gsoc/rubric.html and is nowhere to be found in the repository.

NDOWAH:
Hi, i would like to work on this issue.

PR #1367, opened 2022-04-19
Added GSOC rubric image as SVG

seraphimstreets:
Added GSOC rubric as SVG in relation to #1343, which is about 20KB less than the image.


programmer290399:
Hey @seraphimstreets !!
I am not sure if we'd want to put the image back into the system as we initially removed all the images from git tracking because they were adding to the size of repacks in git (or we believe that was the reason for the very large `.git` folder). So what we'd ideally want to do here is not track any images from now on. 

So we need to look for a reliable image hosting mechanism, what we initially thought was to use GitHub comments and link to those images directly. This is a reliable idea but sort of hacky. If you have any better ideas to host images for free and reliably, please consider sharing.

programmer290399:
But this may not be true for an SVG, as it is more or less like code and not a packed binary file. Thus I'd recommend you to take a look into this and post what you find here, other than that this looks great and shouldn't be an issue for merging.

seraphimstreets:
Oh, so I guess something like adding pictures to Github issues and linking to them? I usually just use Imgur for image hosting, but this seems like a useful hack as well!

Issue #1319, opened 2022-03-10
source: datasets: Add common time-series datasets 


programmer290399:
## Pain Point

Currently, we do not have any commonly used time-series datasets available in dffml 

## Proposed Solution

Write a dataset source (like we have [iris dataset](https://github.com/intel/dffml/blob/master/dffml/source/dataset/iris.py)) to add the following basic datasets:

### Univariate Datasets 
- [ ] [M4 Forecasting data](https://www.kaggle.com/yogesh94/m4-forecasting-competition-dataset)
- [ ] [Household Electricity Consumption](https://paperswithcode.com/dataset/electricity)
- [ ] [Solar Power Dataset](https://paperswithcode.com/dataset/solar-power)  
### Multivariate Datasets
- [ ] [Household Electric Power Consumption](https://www.kaggle.com/uciml/electric-power-consumption-data-set)
- [ ] [Walmart Sales Data](https://www.kaggle.com/c/walmart-recruiting-store-sales-forecasting/data)
- [ ] [M5 Forecasting data](https://www.kaggle.com/c/m5-forecasting-accuracy/data)

pdxjohnny:
Census data? https://registry.opendata.aws/

TirelessClock:
Hey, @pdxjohnny @programmer290399 I'd like to work on this issue. 
Can we use any dataset, or are there any specific requirements (aside from the ones you have written here)?
Also, could you tell me what exactly we would be doing with these datasets?

programmer290399:
Hey @TirelessClock !! 

This issue is a part of a GSoC project for this year, so I am not sure if this is up for grabs or how it would work, you may solve this issue partially, but before proceeding please clear this out with @pdxjohnny. 

As far as the question about "which datasets are to be used" is concerned, we definitely want the ones we have listed above, I have linked to their respective sources, also take a look at how datasets work in DFFML, see the link to iris dataset above. 

But we are open to any other datasets which are commonly used for benchmarking and research purposes. Before implementing one please make sure you're on the same page with the community members so that you don't end up doing work that we'd not be able to merge into the main branch.

I hope this makes it clear, for any further clarification and queries, join our gitter channel.

TirelessClock:
Hello @programmer290399 ,
Thanks for the reply! Yes, I completely understand that it's up for GSoC and it might not necessarily be up for grabs. I would still like to work on the project to solve for at least partially, and at present am working my way through the iris dataset. 
I could use all the help you could give!

[...]

PR #1380, opened 2022-05-10
source: dataset: implement m4 dataset 

TirelessClock:
I've implemented one of the 12 CSV file datasets within the M4 dataset as per issue #1319 
If approved I would implement the other 11 in the same way. 
Thanks!


TirelessClock:
Hey @programmer290399 @pdxjohnny @mHash1m 
Please review this sometime, so I could implement it for the other datasets too.
Also, in the M4 dataset, the test data and the training data are in two different CSV files. How would we deal with that problem?
Thanks!

PR #1396, opened 2022-06-19
AutoML tuners

seraphimstreets:
Implemented tuners.


.gitignore
```diff
@@ -14,6 +14,8 @@ wheelhouse/
 *.eggs
 *.swp
 *.zip
+*.csv
```
programmer290399, line 17: We would want to remove these changes

.pre-commit-config.yaml
```diff
@@ -1,5 +1,5 @@
-repos:
-  - repo: https://github.com/psf/black
-    rev: 19.10b0
-    hooks:
-      - id: black
+- repo: https://github.com/psf/black
```
programmer290399, line 1: We can do this in a different PR and see why it was pinned to the specific version

Issue #968, opened 2021-01-26
model: automl: Hyperparameter tuning


pdxjohnny:
We want to add proper AutoML to DFFML.

The goal here is to create a model which takes as a property of it's config a set of models to used for hyperparameter tuning. Another property of it's config is the set of models which we should attempt to tune (via the first set).

Default values for these results in using all installed models to try to tune all installed model plugins.

To start, we should define a reduced set of models (not all the ones we have). We'll implement automl supporting only this reduced set. The first phase of this project will be to make sure that at one model can be used to tune hyper parameters of another model.

The next phase will be to tune two models using the same tuning model.

The next phase will be to tune two models, using two models. This amounts to doing the previous phase twice, with a different tuning model the second time.

The next phase will be to go through each model in each model plugin we have and see which ones have issues being tuned using the approach taken in the previous phase. This phase will help us determine which properties or methods we may need to add to models to help them self identify and thereby indicate their requirements for hyper parameter tuning, or maybe their inherent lack of support for it.

The final phase will be to implement hyper parameter tuning for N by N models, after implementing what we found to be gaps in the previous phase.

Due to the shortened GSoC cycle, we may end up not doing all of these phases. Which one we go to will be decided as we approach the selection process.

References:
- https://scikit-learn.org/stable/model_selection.html#model-selection

Future work
- We'll later want to expand this to automate dataset cleanup. This will be more challenging and requires that we have more mature dataset cleanup capabilities implemented first. See

BlitzCoder1:
hey i want to contribute to this can you provide a potential start as in how should i proceed

rajpratyush:
@pdxjohnny I have experience of python and have some working experience of Automl as well (did a guided project on coursera - Automl for computer vision) Furthermore  I am currently a Kaggle contributor as well. can you provide some insights on how to begin working on this issue?

pdxjohnny:
@TejasKarkera10 @rajpratyush Updated description

rajpratyush:
@pdxjohnny we can begin this phase by some basic linear regression followed by ridge and lasso regression then move to other type of algorithems and i would generally go with supervised learning algorithms first

pdxjohnny:
@rajpratyush How you choose to begin is up to you :)

Experimentation with approaches here would be wise. As it would create content which would inform ones proposal.

This issue will remain open for everyone to attempt until the GSoC proposal milestone.

rajpratyush:
[...]

PR #1410, opened 2022-08-05
AutoML model incorporating tune commands.

seraphimstreets:
As part of the second stage of the GSOC AutoML project as defined in #968, this is a preliminary iteration of the AutoML model. The idea is to allow users to provide a dataset and list of models they wish to train, and DFFML's integrated AutoML model will perform training/tuning/scoring to select the best model for the user, abstracting away much of the ML process into an easy-to-use API. The current iteration performs the training and scoring using default hyperparameters, so has not implemented tuning yet. Some discussion by the community will be needed to evaluate the best way for tuning to occur in the AutoML process. (should we have default hyperparameter search spaces for each model, or must it be user-defined?)


mHash1m (changes requested):
Great work so far, Lets continue the review on the weekend.

tuner/bayes_opt_gp/dffml_tuner_bayes_opt_gp/tests/test_regressor_model.py
```diff
@@ -0,0 +1,100 @@
...
+from dffml.source.source import Sources
+from dffml import train, score, predict, run_consoletest
+from dffml.util.asynctestcase import AsyncTestCase
+from dffml.feature.feature import Feature, Features
+from dffml.accuracy import MeanSquaredErrorAccuracy
+from dffml.source.memory import MemorySource, MemorySourceConfig
+
+
+from dffml_model_xgboost.xgbregressor import (
+    XGBRegressorModel,
```
pdxjohnny, line 16: Was this file left in intentionally? Let's double check for other additions like this which might be from directory copy pastin getc.

seraphimstreets: Yes, you're right. I'll remove it.

